Issue #1, opened 2016-04-04
Auto-tag node core PR's with sub-system label


phillipj:
Identifying which sub-system(s) PR's relates to by looking at the files changed. This in turn could be used to automatically add sub-system labels, and possibly /cc the correct people/teams listed in [docs/onboard-extras.md](https://github.com/nodejs/node/blob/master/doc/onboarding-extras.md).

Refs https://github.com/nodejs/node/pull/5710#issuecomment-204793840

Qard:
It might be good to detach the auto-cc from the auto-label so labels added manually could also automatically trigger a bot cc'ing the relevant people.

phillipj:
> It might be good to detach the auto-cc from the auto-label so labels added manually could also automatically trigger a bot cc'ing the relevant people.

Good catch, haven't thought of that before. Thanks 👍

Fishrock123:
closing in favor of #31

PR #6, opened 2016-04-04
Auto label node PR's

phillipj:
Very much a work in progress ATM, just wanted to get this out there for public scrutiny and work on it the next couple of days.

**Goals**
- [x] lists appropriate labels for any PR on `/labels/#PR-NUM`
- [x] adds labels to created PR's

**UPDATE** In the discussion below, we've decided to only resolve `c++` and `test` labels at first, then add more labels in later iterations of this.

Refs #1

### DIFF
--- a/lib/node-labels.js
+++ b/lib/node-labels.js
@@ -1,0 +1,44 @@
+'use strict'
+
+const subSystemLabels = {
+  'c++': /^src\//
+}
+
+const rxTests = /^test\//
+
+function resolveLabels (filepathsChanged) {
+  if (isOnly(rxTests, filepathsChanged)) {
+    return ['test']
+  }
+
+  return matchBySubSystemRegex(filepathsChanged)
+}
+
+function isOnly (regexToMatch, filepathsChanged) {
+  return filepathsChanged.every((filepath) => regexToMatch.test(filepath))
+}
+
+function matchBySubSystemRegex (filepathsChanged) {
+  // by putting matched labels into a map, we avoid duplicate labels
+  const labelsMap = filepathsChanged.reduce((map, filepath) => {
+    const mappedSubSystem = mappedSubSystemForFile(filepath)
+
+    if (mappedSubSystem) {
+      map[mappedSubSystem] = true
+    }
+
+    return map
+  }, {})
+
+  return Object.keys(labelsMap)
+}
+
+function mappedSubSystemForFile (filepath) {
+  return Object.keys(subSystemLabels).find((labelName) => {
+    const rxForLabel = subSystemLabels[labelName]
+
+    return rxForLabel.test(filepath) ? labelName : undefined
+  })
+}
+
+exports.resolveLabels = resolveLabels
--- a/lib/node-repo.js
+++ b/lib/node-repo.js
@@ -1,0 +1,70 @@
+'use strict'
+
+const githubClient = require('./github-client')
+
+const resolveLabels = require('./node-labels').resolveLabels
+
+function resolveLabelsThenUpdatePr (options) {
+  githubClient.pullRequests.getFiles({
+    user: options.owner,
+    repo: options.repo,
+    number: options.prId
+  }, (err, res) => {
+    if (err) {
+      return console.error(`! ${prInfoStr(options)} Error retrieving files`, err)
+    }
+
+    const filepathsChanged = res.map((fileMeta) => fileMeta.filename)
+    updatePrWithLabels(options, resolveLabels(filepathsChanged))
+  })
+}
+
+function updatePrWithLabels (options, labels) {
+  // no need to request github if we didn't resolve any labels
+  if (!labels.length) {
+    return
+  }
+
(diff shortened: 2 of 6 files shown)


Fishrock123:
I don't mind if it over-labels a bit. We can probably do a trial of sorts by just regexing against tag names with some sort of priority, falling back to `lib / src` if there are more than, say, 5.

Fishrock123:
the refactoring bits lgtm

phillipj:
Pushed some improvements. Now it does some very basic analysis on the file paths changed and returns a list  of labels.

Obvious things still missing:

- doesn't handle `src` changes
- changes to `lib/_*.js` and `lib/internal` has to be mapped manually. Currently changes to `lib/_stream_readable.js` would be resolved to a "_stream_readable" label

Fishrock123:
I think it is probably more reliable to just hardcode a bunch of regex's. I'm willing to figure out what those should be.

We should also try to fallback to `lib / src` if there is > 5 subsystem labels

williamkapke:
I need the github client extracted out as you did here in `lib/githubClient.js`. Maybe that can go in a PR of its own?

For now I'll use a copy of that file but won't commit it.

Fishrock123:
@phillipj How about we make this only check against regex's and only use c++ from the start, then we can expand the map.

This will also make sure the bot never uses labels that people willy-nilly add/remove.

phillipj:
@Fishrock123 sure, we could start off with a limited set of labels.

What about keeping the current `test` label functionality in here? Adding `test` when the changeset *only* contains `./test/*` files..

Fishrock123:
@phillipj Also seems good to me, but I think it should be cleaned up a bit to be a bit more portable.. We probably need some to check `isOnly` and other to check `is`.

Another example would be a purely benchmarks change. I don't think we should add that yet, just let it be easier to build on after, if possible. :)

test/node-labels.test.js
```diff
@@ -0,0 +1,60 @@
...
+  ])
+
+  t.same(labels, ['c++'])
+
+  t.end()
+})
+
+//
+// Planned tests to be resolved later
+//
```
phillipj: As these tests were already written, I thought it would of some value to leave them for later

lib/node-labels.js
```diff
@@ -0,0 +1,44 @@
...
+}
+
+const rxTests = /^test\//
+
+function resolveLabels (filepathsChanged) {
+  if (isOnly(rxTests, filepathsChanged)) {
+    return ['test']
+  }
+
+  return matchBySubSystemRegex(filepathsChanged)
```
phillipj: @Fishrock123 you think this would suffice for the first iteration?

server.js
```diff
@@ -39,3 +42,16 @@ if (process.env.SSE_RELAY) {
     }
   }
 }
+
+app.get('/labels/:prId', (req, res) => {
```
Fishrock123: Should probably use github events?

Fishrock123:
@phillipj lgtm minus a nit, although I'm not sure we should deploy this yet. We'll want to have a go-ahead from collaborators with some example in a test repo I think.

If you think we should go straight to deploy though, I'm willing to back that.

phillipj:
@Fishrock123 PTAL

This branch is currently deployed and have done some initial tests in https://github.com/TestOrgPleaseIgnore/node/pulls

lib/node-repo.js
```diff
@@ -0,0 +1,45 @@
...
+function updatePrWithLabels (options, labels) {
+  if (!labels) {
+    return
+  }
+
+  githubClient.issues.edit({
+    user: options.owner,
+    repo: options.repo,
+    number: options.prId,
+    labels: labels
```
Fishrock123: Little worried about this, in a theoretical race would this overwrite the existing labels? e.g. if a collaborator opens a PR that adds labels from the start?

scripts/node-subsystem-label.js
```diff
@@ -0,0 +1,23 @@
+'use strict'
+
+const debug = require('debug')('node_subsystem_label')
+
+const nodeRepo = require('../lib/node-repo')
+
+module.exports = function (app) {
+  app.on('pull_request.opened', handlePrCreated)
+
+  function handlePrCreated (event) {
```
williamkapke: This signature is now `(event, org, repo, sender)` if you want to use them and remove the next few lines.

phillipj: Thanks 👍

phillipj:
@Fishrock123 @williamkapke Thanks for reviewing! Just pushed an update which merges the resolved labels with the PR's existing labels, so that we don't accidentally remove any labels.

If it looks good to you, merge at will.

phillipj:
Forgot to mention I tested it with three PRs (9 - 11) in the test repo https://github.com/TestOrgPleaseIgnore/node/pulls

Fishrock123:
Seems good to me, let's try it imo.

PR #7, opened 2016-04-04
Document the startup procedure

phillipj:
This documents the most basic "setup" needed to get the bot up and running.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,22 @@
 
 What's this?? ATM an experiment to see if a bot could be helpful for the Node.js project, to automate all the things!
 
+## Get it started
+
+It's required to have a [GitHub API token](https://github.com/blog/1509-personal-api-tokens) and [Travis token](https://blog.travis-ci.com/2013-01-28-token-token-token/) before starting the bot.
+
+```bash
+$ GITHUB_TOKEN=X TRAVIS_CI_TOKEN=Z npm start
+```
+
+### Create a GitHub access token
+
+Go to your own profile settings page and [generate a personal access token](https://github.com/settings/tokens/new). After clicking the **Generate token** button you're presented with the token you should put in `$GITHUB_TOKEN` when starting the bot.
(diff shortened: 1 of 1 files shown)


phillipj:
Hm... Just realized the current procedure of polling the Travis API for build status also needs its own token.

williamkapke:
I think ultimately this bot will need many tokens for many services. I'm using a `.env` file and https://www.npmjs.com/package/dotenv to load it.

phillipj:
dotenv is nice. Should we still document which tokens are required and how to generate them in the readme?

williamkapke:
I think an example `.env` file in the readme would do the trick:
```
WEBHOOK_SECRET="hush-hush"
GITHUB_TOKEN="xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
TRAVIS_CI_TOKEN="xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
```


I make use of the `silent: true` option so if there isn't an `.env` file- it doesn't do anything.
```js
require('dotenv').load({ silent: true, path: __dirname + '/.env' })
```

Also, Make sure to add `.env` to the `.gitignore` so there are no accidental commits ;).

README.md
```diff
@@ -2,10 +2,22 @@
...
+
 ## Contributing
 
 Please do, contributions are more than welcome!
 
 ## License
 
-MIT
\ No newline at end of file
+MIT
```
Fishrock123: newline EOF please :)

Fishrock123:
seems good to me

README.md
phillipj: > newline EOF please :)

As far as I can see there's a newline EOF present here...

phillipj:
@williamkapke as you introduce dotenv in #8, I'll skip any of that in this PR

PR #8, opened 2016-04-06
emit github events

williamkapke:
<del>Not tested- but should mostly be correct ;)</del>
EDIT: Testing & working!

I'm using a `.env` file like discussed in https://github.com/nodejs-github-bot/github-bot/pull/7#issuecomment-205525555

<del>You'll probably notice that `github-webhook-handler` is referencing my fork- hopefully @rvagg will [accept my PR](https://github.com/rvagg/github-webhook-handler/pull/19) and that can be replaced soon. If not, I can do it a uglier way.</del> EDIT: Needed to go a different direction in the end anyhow.

PR #15, opened 2016-04-08
Use GitHub Status API

Fishrock123:
Here's a test of a [success status](https://github.com/TestOrgPleaseIgnore/test-github-bot/pull/13), and a [failure status](https://github.com/TestOrgPleaseIgnore/test-github-bot/pull/14).

cc @williamkapke & @phillipj

### DIFF
--- a/lib/pollTravis.js
+++ b/lib/pollTravis.js
@@ -22,7 +22,9 @@ githubClient.authenticate({
   token: process.env.GITHUB_TOKEN
 })
 
-function pollThenComment (owner, repoName, prId) {
+exports.pollThenStatus = pollThenStatus
+
+function pollThenStatus (owner, repoName, prId) {
   const prInfo = prInfoStr({ owner, repoName, prId })
 
   // we have to figure out what type of Travis polling we should perform,
@@ -37,7 +39,7 @@ function pollThenComment (owner, repoName, prId) {
     if (hasAnyPrBuilds) {
       pollByPrThenComment(owner, repoName, prId)
     } else {
-      pollByCommitThenComment(owner, repoName, prId)
+      pollByCommitThenStatus(owner, repoName, prId)
     }
   })
 }
@@ -92,7 +94,7 @@ function pollByPrThenComment (owner, repoName, prId, checkNumber) {
  *
  * This is the case for readable-stream.
  */
-function pollByCommitThenComment (owner, repoName, prId) {
+function pollByCommitThenStatus (owner, repoName, prId) {
   const prInfo = prInfoStr({ owner, repoName, prId })
 
   githubClient.pullRequests.getCommits({
@@ -111,7 +113,7 @@ function pollByCommitThenComment (owner, repoName, prId) {
 }
 
 function pollTravisBuildBySha (options, checkNumber) {
-  const createGhComment = createGhCommentFn(options)
+  const createGhStatus = createGhStatusFn(options)
   const prInfo = prInfoStr(options)
   const shaToMatch = options.lastSha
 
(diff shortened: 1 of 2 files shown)


lib/pollTravis.js
```diff
@@ -172,8 +178,27 @@ function createGhCommentFn (options) {
...
+function createGhStatusFn (options) {
+  const prInfo = prInfoStr(options)
+
+  return (state, travisId, message) => {
+    githubClient.statuses.create({
+      user: options.owner,
+      repo: options.repoName,
+      sha: options.lastSha,
+      target_url: `https://travis-ci.org/${options.owner}/${options.repoName}/builds/${travisId}`,
+      context: "Travis CI via nodejs-github-bot",
```
Fishrock123: `nodejs-github-bot/travis-ci` or something may be better, not sure. Not really critical.

lib/pollTravis.js
```diff
@@ -172,8 +178,27 @@ function createGhCommentFn (options) {
...
+      state: state,
+      description: message,
+    }, (err, res) => {
+      if (err) {
+        return console.error(`! ${prInfo} Error while updating GitHub PR status`, err, err.stack)
```
Fishrock123: I changed this to also log the whole `err` since the http errors didn't make stacks, but I can remove that.

lib/pollTravis.js
```diff
@@ -157,7 +163,7 @@ function pollTravisBuildBySha (options, checkNumber) {
 function createGhCommentFn (options) {
   const prInfo = prInfoStr(options)
 
-  return (message, cb) => {
+  return (message) => {
```
Fishrock123: generic cleanup

lib/pollTravis.js
```diff
@@ -34,10 +37,10 @@ function pollThenComment (owner, repoName, prId) {
 
     const hasAnyPrBuilds = res.builds.some((build) => build.pull_request)
 
-    if (hasAnyPrBuilds) {
+    if (false/*hasAnyPrBuilds*/) {
```
Fishrock123: ooops, will correct.

Fishrock123:
Fixes: https://github.com/nodejs-github-bot/github-bot/issues/14

Fishrock123:
rebased

williamkapke:
LGTM

phillipj:
You mentioned on IRC that we should be polling SHA for all repos. Any reason you haven't deleted `pollByPrThenComment()`?

IIRC this means we would do status updates for readable-stream, but still do comments on nodejs.org and citgm.

phillipj:
Have you tested updates on PRs and how that affects previously set statuses? Could we end up with misleading success status on a PR if the first build succeeded, but new commits/pushed rebase results in brokens builds?

Fishrock123:
@phillipj I left that because I wasn't sure if it was still necessary.

Afaik statuses tie to commits, so if the commits are updated GitHub automatically resets it.

Fishrock123:
I'll give it a test when I get home later though.

lib/pollTravis.js
```diff
@@ -138,11 +140,14 @@ function pollTravisBuildBySha (options, checkNumber) {
...
 
       if (lastState === 'passed') {
-        return createGhComment(`[Travis build passed](https://travis-ci.org/${options.owner}/${options.repoName}/builds/${lastBuildForCommit.id}) :+1:`)
+        return createGhStatus('success', lastBuildForCommit.id, 'all tests passed')
       } else if (lastState === 'failed') {
-        return createGhComment(`[Travis build failed](https://travis-ci.org/${options.owner}/${options.repoName}/builds/${lastBuildForCommit.id}) :-1:`)
+        return createGhStatus('failure', lastBuildForCommit.id, 'build failure')
       } else if (~['created', 'started'].indexOf(lastState)) {
         console.log(`* ${prInfo} "${lastState}" build found, will do check #${checkNumber + 1} in 30 seconds`)
+        if (checkNumber === 1) {
```
phillipj: Pretty sure I've seen the first poll resulted in no Travis build found yet, so this check might not always be correct. Maybe we should report "in progress" regardless of how many polls has been made so far?

Fishrock123: Hmm, sure. I didn't because it was a bit spammy for the logs, but should be fine.

williamkapke: I've noticed the same as @phillipj, the first one is often a miss- and line 147 will log on the first check. That will cause it to never show the `pending` state.

Fishrock123:
Hmmm, looks like this no-longer works after being rebased on #8. Investigating.

Fishrock123:
Got it working. My bad.

Looks like this doesn't actually handle PR updates, working on it. (Example: https://github.com/TestOrgPleaseIgnore/test-github-bot/pull/17)

Fishrock123:
Updated, now works for PR updates: https://github.com/TestOrgPleaseIgnore/test-github-bot/pull/17

Fishrock123:
going to open a new PR for some of the other bits

PR #17, opened 2016-04-10
Log when we've got requests with invalid secrets

phillipj:
/cc @Fishrock123 @williamkapke


lib/github-events.js
```diff
@@ -19,6 +19,7 @@ module.exports = (app) => {
     const signature = req.headers['x-hub-signature']
     if (!signature || signature !== sign(secret, req.raw)) {
       res.writeHead(401, 'Invalid Signature')
+      console.error('! event@%s: %s - invalid signature, returning 401', source, data.action)
```
williamkapke: `source` is defined below, `data.action` will not have the event added to it- but it's not as big of a deal

PR #19, opened 2016-04-10
Connect to SSE server when SSE_RELAY set

williamkapke:
I [threw together a relay server](https://github.com/williamkapke/hook-relay) to help with developing the bot. The way it works is you set a `SSE_RELAY` env variable to the url of the relay server that will relay you the events from Github. This removes the need for devs to setup a webhook & routing that points to their local machine... or even worse: deploying to their host, over & over again!

### Overview:
SSE Workflow
**EDIT**: In the picture above, the repos would likely be `TestOrgPleaseIgnore` not `nodejs`.

### Check it out for yourself:
[SSE is a very simple/lightweight protocol](https://html.spec.whatwg.org/multipage/comms.html#server-sent-events) that is perfect for this. It even is supported by modern browsers so you can connect and see the raw event stream.
1) Open: https://hook-relay.herokuapp.com/
2) Go to: https://github.com/TestOrgPleaseIgnore/nodejs.org/pull/1
3) Add a label or a comment...or whatever to cause an event.
4) Watch the output in the browser window.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,26 +1,54 @@
 # Node.js GitHub Bot
 
-What's this?? ATM an experiment to see if a bot could be helpful for the Node.js project, to automate all the things!
-
-## Get it started
-
-It's required to have a [GitHub API token](https://github.com/blog/1509-personal-api-tokens) and [Travis token](https://blog.travis-ci.com/2013-01-28-token-token-token/) before starting the bot.
-
-```bash
-$ GITHUB_TOKEN=X TRAVIS_CI_TOKEN=Z npm start
-```
-
-### Create a GitHub access token
-
-Go to your own profile settings page and [generate a personal access token](https://github.com/settings/tokens/new). After clicking the **Generate token** button you're presented with the token you should put in `$GITHUB_TOKEN` when starting the bot.
-
-### Retrieving your Travis token
-
-Your Travis token is visible on [your profile](https://travis-ci.org/profile) page, by clicking the "show token" link.
+The Node.js Foundation's uses this bot to help manage [the repositories of the GitHub organization](https://github.com/nodejs). 
+It executes [scripts](https://github.com/nodejs-github-bot/github-bot/tree/master/scripts) in response to events that are 
+pushed to it via GitHub webhooks. All [repositories](https://github.com/nodejs) use this same bot instance and need to 
+have the same webhook url & secret configured. Org-wide webhooks are not allow.
 
 ## Contributing
 
 Please do, contributions are more than welcome!
+
+### Environment Variables
+
+- **`GITHUB_TOKEN`**<br>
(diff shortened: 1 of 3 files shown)


Fishrock123:
I don't really understand, what's the benefits of doing this? Seems like an unnecessary middle server?

williamkapke:
@Fishrock123 as mentioned above:
> This removes the need for devs to setup a webhook & routing that points to their local machine... or even worse: deploying to their host, over & over again!

phillipj:
Sorry for this late response on this, been focused on getting build statuses up and running.

Yes, there's absolutely some pain when trying out code changes atm. I'm using dokku, and even as easy as it is to push commits to a remote server, it feels time consuming and strange.

Looking back on the last couple of days, this would probably have made debugging far easier. And you wouldn't have been dependant of me having to redeploy the bot just to test some monkey patching.

I'm leaning towards using the relay server you have created when we actually need it for debugging/trying out new things, rather than always having it as the primary webhook receiver for the repos we use the bot on. Primarly cause it complicates things and really isn't needed when things have settled.

Fishrock123:
Agree with @phillipj

williamkapke:
@phillipj Yes, the relay is for development only. The real nodejs repos would never push to a relay- just directly to the bot server.

I have changes to the readme that describe _developing_ with the relay. I hope to finish it up today and have it added to this PR for review.

williamkapke:
README changes added.

README.md
```diff
@@ -1,26 +1,54 @@
 # Node.js GitHub Bot
 
-What's this?? ATM an experiment to see if a bot could be helpful for the Node.js project, to automate all the things!
+The Node.js Foundation's uses this bot to help manage [the repositories of the GitHub organization](https://github.com/nodejs). 
+It executes [scripts](https://github.com/nodejs-github-bot/github-bot/tree/master/scripts) in response to events that are 
+pushed to it via GitHub webhooks. All [repositories](https://github.com/nodejs) use this same bot instance and need to
```
phillipj: "All repositories use this same bot instance" might give the impression that all repos have bot enabled. Could we be more explicit about us having to enable the bot per repo?

Something like "All repositories *who has the bot enabled*, uses the same bot instance"

phillipj:
> README changes added.

Very nice :+1: Just a little nitpick from me, other than that this LGTM

williamkapke:
@phillipj I reworded that part- I think it is more clear now.

I also added a little part about the `glob` parameter from #25 .

phillipj:
Nice 👍 Looking forward to testing this!

Fishrock123:
@williamkapke do I need to add github webhooks to your SSE relay for this?

williamkapke:
🎉 Cool.

@Fishrock123 It is only for testing/development, so only TestOrgPleaseIgnore needs it- which it already does.

PR #21, opened 2016-04-12
Find commit by PR when polling Travis builds

phillipj:
We've been struggling with PR builds being triggered with commit SHA we have no reference of - seems like some kind of meta merge commit made by GitHub. Therefore our effort of matching builds by the actual commit SHA from PRs doesn't always work.

There's a trivial change included here in 05af2ed which also tries to match commits by their related PR. Fingers crossed this is what we need.

+ reverting the revert (deleting the commenting code)

P.S. this is currently deployed to production.

### DIFF
--- a/lib/pollTravis.js
+++ b/lib/pollTravis.js
@@ -22,77 +22,10 @@ githubClient.authenticate({
   token: process.env.GITHUB_TOKEN
 })
 
-exports.pollThenStatus = pollThenStatus
-
-function pollThenStatus (owner, repoName, prId) {
-  const prInfo = prInfoStr({ owner, repoName, prId })
-
-  // we have to figure out what type of Travis polling we should perform,
-  // either by PR #id (as for nodejs.org) or commit sha (for readable-stream)
-  travisClient.repos(owner, repoName).builds.get((err, res) => {
-    if (err) {
-      return console.error(`! ${prInfo} Error while retrieving initial builds`, err.stack)
-    }
-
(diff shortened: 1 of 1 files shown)


lib/pollTravis.js
```diff
@@ -129,7 +63,9 @@ function pollTravisBuildBySha (options, checkNumber) {
       return console.error(`! ${prInfo} Got error when retrieving Travis builds`, err.stack)
     }
 
-    const matchingCommit = res.commits.find((commit) => commit.sha === shaToMatch)
+    const matchingCommit = res.commits.find((commit) => {
+      return commit.sha === shaToMatch || commit.pull_request_number === prToMatch
```
phillipj: This is the change I'm trying to see if does the trick

phillipj:
Tested with a branch from the main repo https://github.com/nodejs/nodejs.org/pull/490. We need a test PR from a nodejs.org fork as well.

Fishrock123:
See https://github.com/nodejs/nodejs.org/pull/647#issuecomment-208940672 -- this sometimes incorrectly gets outdated builds

Fishrock123:
Hmm, looks like it's working now. Strange.

lib/pollTravis.js
Fishrock123: Hmm, I think we somehow need to ensure we prioritize matching the last commit in a pr to avoid races.

Fishrock123:
@phillipj might as well merge for now?

phillipj:
@Fishrock123 sure enough

Issue #22, opened 2016-04-12
avoid race conditions, only poll by last commit of a PR


Fishrock123:
See https://github.com/nodejs-github-bot/github-bot/pull/21#discussion_r59326377

phillipj:
This should be fixed with #24 merged.

PR #24, opened 2016-04-12
Also match commits by their commited date when polling Travis

phillipj:
As seen in the nodejs.org repo today https://github.com/nodejs/nodejs.org/pull/647#issuecomment-208940672, the bot might be tripped into reporting status for the previous build/commit.

This aims to fix that issue by always matching commits by their date as well as commit sha or PR number.

Refs #22


phillipj:
Even seems to be working :dancers:

PR #25, opened 2016-04-12
Use globs to load scripts

williamkapke:
The (current) initial version just tries to load all files in the `./scripts` directory. When developing, you want to target just the script you're working on.

This change allows a `glob` argument to be passed in to specify which file(s) to try to load. If it isn't specified, it defaults to `./scripts/**/*.js`.


phillipj:
:+1:

PR #28, opened 2016-04-15
move sse code to github-events

williamkapke:
Having the events emitting in 2 places is not good. The SSE relay is only `github-events` related code - so it makes sense to move it to the `github-events.js` file.

Also, I am now emitting the org, repo, and sender names since will be needed by most listeners.


Fishrock123:
seems good to me

Issue #31, opened 2016-04-20
Node core auto labeling improvements


phillipj:
## Low hanging fruits

- [x] `doc/` for doc
- [x] `benchmark/` for benchmark
- [x] `deps/$name` -> $name (libuv, v8, etc)
- [x] `configure`, `*Makefile`, `node.gyp`, `common.gypi` -> `build` (a lot of stuff in tools/ lives here too but not necessarily low hanging fruit)

*To avoid duplicate work on these, please leave a comment if you'd like to give some of these a shot.*

## Challenges

Regarding the `c++` label for any PR that has file changes in `src/`:

> For example, there might be a PR that makes changes to both `lib/net.js` and `src/tcp_wrap.cc`, but the overall net change in `lib/net.js` is substantially greater than that of `src/tcp_wrap.cc`. In that particular case I would normally just tag it with `net` only.

## How to work on this

Its current labels brain is located in [lib/node-labels.js#resolveLabels()](https://github.com/nodejs-github-bot/github-bot/blob/master/lib/node-labels.js#L9). It's even covered by tests [test/node-labels.test.js](https://github.com/nodejs-github-bot/github-bot/blob/master/test/node-labels.test.js).

Refs nodejs/node#6247

Fishrock123:
We may want to blacklist `node_version.h` from the `c++` label if possible: https://github.com/nodejs/node/pull/6383#event-640812288

jasnell:
I've been considering adding a `release` label that we can use to tag release proposals. Would it be easier to key off that instead of blacklisting `node_version.h`?

phillipj:
@Fishrock123 @jasnell fixed it in https://github.com/nodejs-github-bot/github-bot/pull/32 by ignoring `src/node_version.h`

phillipj:
Closing this for now as the originally mentioned labels has been fixed.

PR #32, opened 2016-04-22
Added doc, benchmark and deps/X node core PR labels

phillipj:
This improves the current node core auto PR labelling functionality by supporting these extra labels:

- `doc` for exclusive `doc/*` changes
- `benchmark` for exclusive `benchmark/*` changes
- `v8`, `libuv`, `openssl` etc when having changes in `deps/X/*`

Also ignores changes to `src/node_version.h` to avoid labelling that as `c++` as mentioned in https://github.com/nodejs-github-bot/github-bot/issues/31#issuecomment-214572696.

@jbergstroem @mscdex @Fishrock123 does these labels sound reasonable to you?

Refs #31, https://github.com/nodejs/node/pull/6247

### DIFF
--- a/lib/node-labels.js
+++ b/lib/node-labels.js
@@ -1,27 +1,36 @@
 'use strict'
 
-const subSystemLabels = {
+const subSystemLabelsMap = {
   'c++': /^src\//
 }
 
-const rxTests = /^test\//
+const exclusiveLabelsMap = {
+  test: /^test\//,
+  doc: /^doc\//,
+  benchmark: /^benchmark\//
+}
 
 function resolveLabels (filepathsChanged) {
-  if (isOnly(rxTests, filepathsChanged)) {
-    return ['test']
-  }
+  const exclusiveLabels = matchExclusiveSubSystem(filepathsChanged)
 
-  return matchBySubSystemRegex(filepathsChanged)
+  return (exclusiveLabels.length > 0)
+          ? exclusiveLabels
+          : matchAllSubSystem(filepathsChanged)
 }
 
-function isOnly (regexToMatch, filepathsChanged) {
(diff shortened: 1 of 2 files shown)


phillipj:
I'm merging this as no objections has been raised.

williamkapke:
post LGTM! ;)

phillipj:
@williamkapke thanks! 😄

lib/node-labels.js
```diff
@@ -1,27 +1,42 @@
 'use strict'
 
-const subSystemLabels = {
-  'c++': /^src\//
+// order of entries in this map *does* matter for the resolved labels
+const subSystemLabelsMap = {
+  // don't want to label it a c++ update when we're "only" bumping the Node.js version
+  'c++': /^src\/(?!node_version\.h)/,
```
Fishrock123: Nice!

Fishrock123:
belated LGTM. I'll try to find time to tackle some of the more complex ones soon

Fishrock123:
https://github.com/nodejs/node/pull/6432 Seems to have been labeled as `test` when it doesn't only contain tests. Investigating.

Fishrock123:
Hmm, seems like this doesn't *only* label `test`, `doc`, `benchmark` if it only contains those files, but rather if it is only one of those labels. I'll try to fix tomorrow.

phillipj:
> Hmm, seems like this doesn't only label  `test`, `doc`, `benchmark` if it only contains those files, but rather if it is only one of those labels. I'll try to fix tomorrow.

Woops, cause the old `isOnly()` got removed and we're obviously missing a test for that behavior.

@Fishrock123 goodie! I'll be ready to deploy as soon as you've got a fix ready.

Issue #5, opened 2016-04-04
Monitor updates to pr and post Jenkins CI progress


jbergstroem:
Background and further details here: https://github.com/nodejs/build/issues/236

I'd be happy to chip in here.

williamkapke:
@phillipj Looking at @jbergstroem's work for the build team in that linked issue above, I see he/they are using the [Github Status API](https://developer.github.com/v3/repos/statuses/). That seems like a cool way to go... maybe incorporate that in to the work you have in pollTravis?

Fishrock123:
Yeah we should probably push to the github status API. That way, it also displays at the merge button and you don't need to find the latest comment.

Fishrock123:
Related `github` module API: http://mikedeboer.github.io/node-github/#statuses, looking into it.

phillipj:
@williamkapke trying to use to status API sounds awesome! One step closer to mimic Travis completely :dancers:

Fishrock123:
moving the status api discussion to https://github.com/nodejs-github-bot/github-bot/issues/14

phillipj:
@jbergstroem how would we know *when* to start polling? Could Jenkins do a HTTP requestagainst the bot, notifying there's a PR build that just started? Or should the bot listen for PR comments matching "CI: http://ci.nodejs.org/*"?

jbergstroem:
I owe a few of you a post -- sorry for the delay. I'm a bit backlogged at the moment; will get back to you shortly.

phillipj:
Seems the `../api/json` endpoint would be perfect for polling build status. E.g. for https://github.com/nodejs/node/pull/6342: https://ci.nodejs.org/job/node-test-pull-request/2363/api/json

phillipj:
Landed with #35

Issue #4, opened 2016-04-04
Write an audit log


jbergstroem:
Write a log of all commits, opening/closing pr's, etc so we can track whether malicious commits or force pushes were made in case of an incident.

williamkapke:
👍  This is something I was really pondering. Also- public visibility? I would lean towards yes- but I guess it depends on what is logged. For an example, see the thoughts here: https://github.com/nodejs/inclusivity/issues/85#issuecomment-204236281

jbergstroem:
I agree with public visibility when we've had the chance to look at output ourselves. This would involve things like changes to memberships too. We likely need to look at all data and draw a line.

jbergstroem:
As for making this persistent I guess the choice lies with the person that opens the PR, but personally I'd be keen on using postgres over stuff like mongodb.

phillipj:
+1 for postgres. Worth mentioning I think the most important thing is how easy and well suited the db is for querying for insights later, not necessarily if it's easy to insert data.

Trott:
This is more than 8 years idle and we've gotten by with the GitHub audit log, so I'm going to close this. But if someone feels like it should remain open or is interested in doing the work, please re-open, comment, or open a new issue.

PR #35, opened 2016-04-28
Proxy Jenkins build status to GitHub

phillipj:
This is work in progress atm, PR exists for public scrutiny and to show I've started working on it.

Any thoughts are much appreciated!

Refs #5


Fishrock123:
Is this page the only official docs on it? https://wiki.jenkins-ci.org/display/JENKINS/Remote+access+API

cc @jbergstroem

phillipj:
Made some improvements to this today. My biggest question atm is how we should trigger Jenkins polling in the first place?

I've thought of two different approaches so far:

1. Does Jenkins have support for notifying the bot with a HTTP request? If so that could be a pretty clean approach.
2. Listen for PR comments matching "CI: https://ci.nodejs.org/job/node-test-pull-request/X".

Any other thoughts?

/cc @Fishrock123 @jbergstroem

Fishrock123:
All of my existing ideas have been around suggestion 1. -- by far the most reliable.

jbergstroem:
@phillipj we could implement it [pinging the bot] as a build step. The problem here is security; the receiving end can't use much else than ip/hostname matching for security. If that's acceptable it would likely be the best path forward.

As for methodology; I'd like to suggest the following:
1. (based on option one) accept the incoming request
2. inspect what subjobs are being spawned
3. post a 'pending' request for all subjobs to the PR
4. wait for subjobs to complete
5. add a completion task at each job which posts info to bot (again, if trust can be established)
6. parse info and notify github. I'd really like to see something like:
   1. if test passes, something basic (ci/linux: all tests passed)
   2. if we get flakey/skips, use pass again but mention flakey/skips (ci/linux: tests passed (5 flakey, 5 skips))
   3. if we get fails, post as failure with num fail tests (ci/linux: 2 failed tests)

Also, once we close in on this I reckon we should split linter from the test-pr job and start calling it separately from the gh bot on pretty much every commit, as well as create a set of slaves that does the same for commit message validation.

Fishrock123:
> The problem here is security; the receiving end can't use much else than ip/hostname matching for security.

Can it be over https? Github and the like just send a secret to verify, or the hash of a secret. Would that be possible?

That being said ip/hostname matching probably isn't too bad?

--

I like the other methodology suggestion. @jbergstroem how difficult would it be to make those subjobs so we can start trying this?

jbergstroem:
@Fishrock123 the problem isn't so much man in the middle, but the fact that we can't store credentials at jenkins since a PR could potentially just do `echo $FOO`. I actually implemented posting progress to github as part of the jenkins jobs but abandoned due to security issues. More info about findings over here https://github.com/nodejs/build/issues/236.

@Fishrock123 by subjobs i was referring to the fact that `node-test-pr` invokes `node-test-commit` which in turn invokes `node-test-linux` and so on. So what we'd do is likely send a request from each (if `POST_STATUS_TO_PR` is set at top level) at start and completion.

phillipj:
@jbergstroem no problem going for ip/host matching for incoming HTTP requests.

To ensure we're on the same page here: if the bot gets HTTP requests upon subjob start and completion, it will not perform any Jenkins polling at all, but will work more like a proxy for updating github PR (commit) status? ..if so that sounds very good :+1:

It would be very helpful if we could get our hands on some example payloads those subjob requests from Jenkins include.

Fishrock123:
> if the bot gets HTTP requests upon subjob start and completion, it will not perform any Jenkins polling at all, but will work more like a proxy for updating github PR (commit) status?

@phillipj correct.

> It would be very helpful if we could get our hands on some example payloads those subjob requests from Jenkins include.

Yes. this. :D

jbergstroem:
Example payloads: I can craft them whatever way we want -- will essentially just pass stuff through curl.

phillipj:
@jbergstroem we should come a long way with:

- build parameters
- subjob name
- subjob URL
- commit sha
- build result (upon completion)

The structure of the payload isn't important, as long as it's JSON. Is it doable?

Fishrock123:
What do we need build params for?

Fishrock123:
Imo, we'll need:

- name
- url
- commit sha
- status

phillipj:
> What do we need build params for?

@Fishrock123 the POST_STATUS_TO_PR option. Should we simply ignore it and build 'em all?

Fishrock123:
@phillipj I presume that would determine if the bot gets contacted or not?

phillipj:
@Fishrock123 ahh that makes sense. Agreed we could do wo/build params then.

jbergstroem:
More a summary of yours, but yeah; this is what I'm thinking:

##### on start:
 - the jenkins job will check if `POST_STATUS_TO_PR` is set or bail
 - curl to $url/$start_endpoint with a pretty low timeout since this is sync, with:
   - job name (we can map job name to something more user friendly at the gihub bot, for instance: node-test-linux -> test/linux)
   - url to job
   - commit sha
   - designated slave

##### on completion:
 - the jenkins job will check if `POST_STATUS_TO_PR` is set or bail
 - curl to $url/$stop_endpoint
   - job name
   - url to job
   - commit sha
   - designated slave
   - build results (this will be different between some jobs; for instance linter or commit-message-validator):
     - passed/flaky/fail
     - total tests/flaky/failed
     - time spent

We should poll jobs occasionally slave might go away during job. Not very likely but it happens. Also, we sometimes have ci-related failures. Reporting them as a "ci fail" would be a nice ui improvement.

phillipj:
@jbergstroem @Fishrock123 just did some adjustments based on what we've been discussing here.

I did some assumptions on the structure and made three fixtures:

- [test/_fixtures/pending-payload.json](https://github.com/nodejs-github-bot/github-bot/blob/pull-jenkins/test/_fixtures/pending-payload.json)
- [test/_fixtures/success-payload.json](https://github.com/nodejs-github-bot/github-bot/blob/pull-jenkins/test/_fixtures/success-payload.json)
- [test/_fixtures/error-payload.json](https://github.com/nodejs-github-bot/github-bot/blob/pull-jenkins/test/_fixtures/error-payload.json)

You can see the result of POSTing those payloads in https://github.com/TestOrgPleaseIgnore/node/pull/13

What's missing from @jbergstroem last comment is use of *designated slave* and *different kinds of build results*. It also doesn't use two different endpoints for start / completion, I did see the need for that so far.

..and it would have been nice to get this covered by tests.

test/_fixtures/error-payload.json
```diff
@@ -0,0 +1,12 @@
+{
+    "fullDisplayName": "node-test-commit-arm #3087",
+    "result": "FAILURE",
```
Fishrock123: maybe `status`

test/_fixtures/error-payload.json
```diff
@@ -0,0 +1,12 @@
+{
+    "fullDisplayName": "node-test-commit-arm #3087",
```
Fishrock123: `info`?

test/_fixtures/error-payload.json
```diff
@@ -0,0 +1,12 @@
...
+    "result": "FAILURE",
+    "changeSet": {
+        "items": [
+            {
+                "commitId": "7190bd23810f0880d95605bf6f785224a345d477"
+            }
+        ]
+    },
```
Fishrock123: Mind explaining the rational behind this?

I can't really see the use of anything other than one commit sha

lib/push-jenkins-update.js
```diff
@@ -0,0 +1,80 @@
...
+    state: 'error',
+    message: 'build aborted'
+  },
+  'UNSTABLE': {
+    state: 'error',
+    message: 'build unstable'
+  }
+}
+
+const rxDisplayName = /node\-test\-commit\-(\w+)/
```
Fishrock123: I don't think the linter has this name, is there that much use in regexing this?

lib/push-jenkins-update.js
```diff
@@ -0,0 +1,80 @@
...
+    state: 'failure',
+    message: 'build failure'
+  },
+  'ABORTED': {
+    state: 'error',
+    message: 'build aborted'
+  },
+  'UNSTABLE': {
+    state: 'error',
+    message: 'build unstable'
```
Fishrock123: is this flaky tests?

lib/push-jenkins-update.js
jbergstroem: We don't follow a strict naming convention for jobs; I'd even consider mapping jobs to friendlier names here.

jbergstroem: ..or we just trust naming sent from each job?

lib/push-jenkins-update.js
```diff
@@ -0,0 +1,80 @@
...
+const rxDisplayName = /node\-test\-commit\-(\w+)/
+
+function pushJenkinsUpdate (options, build) {
+  const statusOpts = extendWithBuildData(options, build)
+  const createGhStatus = createGhStatusFn(statusOpts)
+  const isPending = build.result === null
+  const matchedGhStatus = jenkinsToGhStatusMap[build.result]
+
+  if (isPending) {
+    createGhStatus('pending', 'build in progress')
```
jbergstroem: I'd like this to be a bit more flexible. We could even post updates between compiling and testing. Linters aren't necessarily "building" either. Perhaps having a set of defaults or pass on what is posted through curl?

phillipj:
Thanks for reviewing so quick guys!

> Mind explaining the rational behind this?

@Fishrock123 it's the same structure as the Jenkins json/api uses. If we're not tied to that structure/format and can do adhoc, I'm more than happy to keep it simpler.

lib/push-jenkins-update.js
phillipj: Sure, we could trust whatever curl pushes, but keep in mind the state (first param) has to be one of `pending|success|error|failure`.

Some explanation of what's provided when invoking `githubClient.statuses.create(..)` to set the github status and what's displayed on github.com:

- **state**: what decides the icon/color to the left
- **context**: text in the middle in bold
- **description**: plain text to the right

image

test/_fixtures/error-payload.json
phillipj: We really just need the last commit, as the changes in previous commits would be included in the last commit because of git's parent refs. This is also visible in what Travis does to PRs, only the last commit included in the PR will get the "build state icon".

lib/push-jenkins-update.js
phillipj: > ..or we just trust naming sent from each job?

Totally fine with me 👍

lib/push-jenkins-update.js
phillipj: I've found these statuses by looking at the [build trend](https://ci.nodejs.org/job/node-test-pull-request/buildTimeTrend), this was marked `UNSTABLE`: https://ci.nodejs.org/job/node-test-binary-windows/1866/

phillipj:
As just mentioned ^ if we're able to use whatever structure/format we want, just ignore the fixtures added here.

@jbergstroem would you be able to post an example of a what these curl jobs could provide the bot with?

lib/push-jenkins-update.js
jbergstroem: Perhaps I'm overthinking it but we could even use variations for the avatar, for instance operating system/identifier or type of job? (test, lint, commit message validator, ...)

test/_fixtures/error-payload.json
Fishrock123: @phillipj should we just have `commit` then?

lib/push-jenkins-update.js
Fishrock123: Means more github accounts then, right?

test/_fixtures/error-payload.json
jbergstroem: Its really up to us how we want to handle it -- we could test every commit if we wanted to -- but since this will be driven by manual triggers (someone creates a test-pr job) we don't really care what commit it is, just posting an update against it.

lib/push-jenkins-update.js
jbergstroem: @Fishrock123 I think you can provide an avatar per job status update

jbergstroem:
@phillipj jenkins basically has the same information available as what you can get from the api; so I'll just build a query similar to what we expect. I will build different queries based on fail/success/etc.

Fishrock123:
@jbergstroem How about making it curl an object like this, and we'll go from there?

```json
{
  "info": "something about node-test-commit-linux",
  "status", "FAILURE",
  "commit": "shasum",
  "url": "https://ci.nodejs.org/job/node-test-commit-linux/3176/"
}
```

jbergstroem:
@Fishrock123 sounds good. in addition to that, we should probably add `identifier` (linux, linter, etc) and test results. I'll look at my old stuff.

jbergstroem:
How about something like:
```bash
$ curl \
  --connect-timeout 5 \
  -X POST \
  -H "Content-Type: application/json" \
  -d "{ \
    \"identifier\": \"linter\", \
    \"status\": \"pending\", \
    \"url\": \"${BUILD_URL}\", \
    \"commit\": \"$(git rev-parse HEAD)\", \
    \"message\": \"checking for errors\" \
  }" \
  https://deepthought.nodejs.org/endpoint/
```

..and with payload:
```bash
curl \
  --connect-timeout 5 \
  -X POST \
  -H "Content-Type: application/json" \
  -d "{ \
    \"identifier\": \"test/linux\", \
    \"status\": \"failure\", \
    \"url\": \"${BUILD_URL}\", \
    \"commit\": \"$(git rev-parse HEAD)\", \
    \"message\": \"tests failed\",
    \"data\": { \
      \"total\": \"12345\", \
      \"pass\": \"12340\", \
      \"fail\": \"3\", \
      \"skip\": \"2\" \
    } \
  }" \
  https://deepthought.nodejs.org/endpoint/
```

The rationale behind abstracting data is that we could store it and start generating insights over time. I see something similar as part of the other data we collect from github by receiving all the events (briefly mentioned in #4). It's also easier working with the data -- `one test failed (5 skipped, 2 flakey)` vs `2 tests failed ..` than through bash.

phillipj:
@jbergstroem LGTM, I'll update the fixtures tonight. I suggest we leave `data` out of the picture initially to get this out asap, and add it later in a separate PR which also stores that data somehow.

lib/push-jenkins-update.js
phillipj: Can't find any reference saying we can provide avatar per status update;

- https://developer.github.com/v3/repos/statuses/#parameters
- https://mikedeboer.github.io/node-github/#statuses.prototype.create

jbergstroem: You're right; I just read the wrong docs.

jbergstroem:
@phillipj that means I'll be formatting the failures from bash then?

phillipj:
@jbergstroem hm... you would prefer the status description (the text displayed to the right of the status) to be a combination of curl's `message` and a human readable format of `data`?

Fishrock123:
@jbergstroem I think you can include the data but we will worry about processing after we have the initial things up and running?

phillipj:
Just updated the fixtures, and deleted lots of code which ended up giving a lot of flexibility from the curl side of things. It's now very much a proxy to update GH statuses, forwarding whatever it gets as input, rather than trying to do smart transformations.

PTAL and shout if I've missed something from your comments.

image
https://github.com/TestOrgPleaseIgnore/node/pull/14

jbergstroem:
@Fishrock123 yeah I agree. So, where do I post for now? Also, I'm kind of interested in finding a way of queueing/delegating all the curl requests so they return instantly -- anyone got any ideas?

Fishrock123:
http://github-bot.nodejs.no/hooks/github I think

phillipj:
@Fishrock123 nope, that's for the github repo hooks (used to trigger Travis polling). I'll send the URL to @jbergstroem on IRC as the location of the bot shouldn't be public atm since we don't have a ip/hostname/other security measures in place.

phillipj:
@Fishrock123 @jbergstroem soooo, I just pushed a commit loading the gun so to speak ^.

ATM the bot gets curl updates from node-test-linter in production, but doesn't do anything since these changes hasn't been merged yet. Are we ready to merge & deploy this as is, or do we need more things before rolling forward?

Fishrock123:
I've made a test PR at https://github.com/nodejs/node/pull/6674 -- forgot to make this on the repot itself though. LMK if it should open a new one from a node repo branch so you can push to it.

phillipj:
..and now it even has tests ensuring github is requested with the expected payload as well 🎆

phillipj:
@jbergstroem fyi I'm holding this 'til you respond. I could potentially deploy this specific branch instead of merging it to master, if you'd rather do a couple of test runs before we going full throttle ;)

jbergstroem:
I've now implemented check for POST_STATUS_FOR_PR. I'll default that to off for now and start adding status updates to more jobs.

phillipj:
This is now live in production 💥

PR #38, opened 2016-05-09
Created smoketest verifying the bot starts

phillipj:
As a means of being more confident deploying the bot automatically to production upon merge to master, this ensures the bot at least starts as planned and responds on a trivial GET request.

Did some refactoring to make the bot more testable, most noticeably extracted the `server.listen()` part into it's own file so the `app` could be tested in isolation with explicit control over the server/listening part for testing purposes. Plus not always starting the SSE relay client (which we don't need while testing).

Refs https://github.com/nodejs/build/issues/404.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,3 +3,4 @@ language: node_js
 node_js:
   - 4
   - 5
+  - 6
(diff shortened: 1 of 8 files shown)


Fishrock123:
lgtm

PR #39, opened 2016-05-11
Log job name when proxying Jenkins -> GH status

phillipj:
Currently we're not logging *which* Jenkins job is trying to push GitHub status updates. That information would be very valuable when debugging.

/cc @jbergstroem


phillipj:
As it's a trivial change, I'm merging this as some github status update errors are getting logged where this would be very helpful.

PR #40, opened 2016-05-11
Validate POST payloads from Jenkins

phillipj:
So far there's been cases when sha, or other required fields of the payload has been missing, but the bot has willingly tried to proxy it as well as responding with a misleading 201 status code.

This at least does some trivial validation and tells you right away if the payload is seen invalid.


Fishrock123:
lgtm if tests pass

phillipj:
> lgtm if tests pass

Travis says so :smile_cat:

PR #41, opened 2016-05-12
Refactor labels to use Regex-keyed Maps

Fishrock123:
Allows for more extensibility where labels may be the same. (e.g. more than one `$1` mapping)

Requires Node.js v6.0.0+ for the destructing in the `for of` loop.

cc @phillipj

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,4 @@
 sudo: false
 language: node_js
 node_js:
-  - 4
-  - 5
   - 6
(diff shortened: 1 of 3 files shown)


Fishrock123:
v4 and v5 obviously fail with this

phillipj:
Nice refactoring 👍 

While you're at it, could you remove v4 and v5 from `.travis.yml`, and set engines.node to `">= 6.0.0"` in `package.json`? Otherwise it LGTM.

Fishrock123:
@phillipj done

PR #42, opened 2016-05-12
Expose traceable logs

phillipj:
*Sorry about the big change set but logging cuts through most of the code*

ATM debugging what the bot does and errors isn't trivial for several reasons. We don't log enough about incoming events or Jenkins updates, and we don't have any way to trace whatever happened to one particular request.

It would also be nice to let more people have a look at the logs at any given time. Only having one person with access to them (which is me atm), isn't good enough.

These changes introduces bunyan for logging. Combined with an express middleware it'll add unique request ids to each incoming request, which we use throughout giving us the ability to search for everything that happened for any given request.

All logs are exposed and available for those who know the username / password combination set in `$LOGIN_CREDENTIALS`.


app.js
```diff
@@ -5,16 +5,23 @@ require('dotenv').load({ silent: true })
...
+const bunyanMiddleware = require('bunyan-middleware')
+
+const logger = require('./lib/logger')
 
 const captureRaw = (req, res, buffer) => { req.raw = buffer }
 
 const app = express()
 app.use(bodyParser.json({ verify: captureRaw }))
+// bunyanMiddleware gives us request id's and unique loggers per incoming request,
+// for satefy reasons we don't want to include the webhook GitHub secret in logs
```
Fishrock123: Should we whitelist instead, or is that too impractical?

lib/github-events.js
```diff
@@ -22,22 +22,28 @@ module.exports = (app) => {
 
     if (!signature || signature !== sign(secret, req.raw)) {
       res.writeHead(401, 'Invalid Signature')
-      console.error('! event@%s: %s - invalid signature, returning 401', data.repository.full_name, data.action)
+      req.log.error('Invalid GitHub event signature, returning 401')
```
Fishrock123: where do you get `req.log`?

logs/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
+!.gitignore
```
Fishrock123: newline please

Fishrock123:
Seems fine, but tests aren't passing. I didn't 100% review all the variable passing.

williamkapke:
What about using a service like Papertrail?

lib/github-events.js
phillipj: [bunyan-middleware](https://github.com/tellnes/bunyan-middleware) puts on every incoming request

app.js
phillipj: ATM whitelisting is not implemented https://github.com/tellnes/bunyan-middleware#options. We could ofc open a PR if we really need it. Got any other sensitive headers we should avoid in the logs?

phillipj:
> What about using a service like Papertrail?

I see this as a intermediate step between the `console.log`-ing we're doing today, and something like Papertrail. By using bunyan w/its cli we're able to get a lot of value of a very simple solution. Hopefully the bot will get deployed in the project's infra, idk what's preferred or not there in regards to such 3rd party services, which is also one of the reasons I'd like to keep it local and simpler at first.

lib/logger.js
```diff
@@ -0,0 +1,23 @@
...
+  streams: [
+    {
+      stream: process.stdout,
+      level: stdoutLevel
+    },
+    {
+      type: 'rotating-file',
+      path: path.join(__dirname, '../logs/bot.log'),
+      period: '1d', // daily rotation
+      count: 3      // keep three days of logging
```
jbergstroem: This looks like a very low default; especially if we'd be audit logging. Perhaps even avoid deletion? Optimally I'd like to see this passed as an option as well.

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.0.0-beta1",
   "description": "Node.js GitHub Bot",
   "scripts": {
-    "start": "node server.js",
+    "start": "node server.js | bunyan -o short",
```
jbergstroem: will this work as-is? `bunyan` wouldn't necessarily live in path; perhaps `node_modules/.bin/bunyan` ?

TheAlphaNerd: npm scripts shim in `node_modules/.bin` into the path :smile:

jbergstroem: oh. magic!

phillipj:
Pushed an update which exposes the log files ^.

This completes my initial thoughts about exposing the logs to everyone who knows the user/pass, and maybe do some interesting queries on them:

```bash
# everything
$ curl http://example.org/logs/bot.log | bunyan -o short

# loglevel
$ curl http://example.org/logs/bot.log | bunyan -l error

# filter by PR #
$ curl http://example.org/logs/bot.log | bunyan -c 'this.pr == "nodejs/nodejs.org/#42"'

# filter by unique request
$ curl http://example.org/logs/bot.log | bunyan -c 'this.req_id == "some uuid here"'
```

lib/logger.js
phillipj: Made env variable `$KEEP_LOGS` for configurability, defaults to 10 days.

phillipj:
Merging this as I've addressed issues mentioned.

PR #44, opened 2016-05-15
labels: fix string_decoder labeling

mscdex:
`buffer` is really separate from `string_decoder` (the latter actually uses the former's functionality), so update the bot's labeling mechanism.


phillipj:
LGTM

phillipj:
Deployed. Thanks for jumping in and opening a PR 👍

PR #53, opened 2016-07-02
labels: prevent deps from getting "lib / src" label

### LINKED ISSUE #52: label error for changes in V8 directory
See https://github.com/nodejs/node/pull/7488

The bot added a `lib / src` label but the changes are only in `deps/v8`

phillipj:
This prevents changes in `deps/` from getting labelled `lib / src` as seen in https://github.com/nodejs/node/pull/7488.

@Fishrock123 as you implemented `lib / src` in the first place, does this seem good to you?

Closes #52

### DIFF
--- a/lib/node-labels.js
+++ b/lib/node-labels.js
@@ -104,7 +104,7 @@ function matchSubSystemsByRegex (rxLabelsMap, filepathsChanged, limitLib) {
 
     for (var i = 0; i < mappedSubSystems.length; ++i) {
       const mappedSubSystem = mappedSubSystems[i]
-      if (limitLib && jsSubsystemList.includes(mappedSubSystem)) {
+      if (limitLib && hasJsSubsystemChanges(filepathsChanged, mappedSubSystem)) {
         if (jsLabelCount.length >= 4) {
           for (const jsLabel of jsLabelCount) {
             delete map[jsLabel]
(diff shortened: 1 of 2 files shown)


Fishrock123:
Doesn't seem quite right to me, it should only be checking if there are 4+ `jsLabel`s, not matching these files..?

phillipj:
Hum... On second thought it looks weird to me too, I'll do another shot at this

phillipj:
@Fishrock123 PTAL. 

The primary cause behind what caused https://github.com/nodejs/node/pull/7488 to be labeled `lib / src`, is that it there's 4+ files which is seen to be labeled `v8` since the files live in `deps/v8/`. At the same time `jsSubsystemList.includes(mappedSubSystem)` returns true cause `v8` is also a valid JS subsystem.

While digging into this, I got confused about your recent answer

> it should only be checking if there are 4+ jsLabels, not matching these files..?

Should `lib / src` actually *only* match JS subsystems? If so, why is `src` parth of the label?

The commit I just pushed, double checks if the file changes made *really* is JS subsystem by ensuring the files live in `lib/` - that might not be correct if `src/` files also should be limited?

Fishrock123:
> Should `lib / src` actually *only* match JS subsystems? If so, why is `src` parth of the label?

Ok, maybe that's not entirely true, but it does make it easier for the bot to deal with.

The label is designed for when something touches many subsystems or large parts of the underlying codebase.

Since we don't label C++ files by subsystem yet, going an all-JS route seems the solution for now.

phillipj:
> Since we don't label C++ files by subsystem yet, going an all-JS route seems the solution for now.

@Fishrock123 in other words this PR should be good to go then, since it ensures the affected files live in `./lib/*` before doing any `lib / src` checks.

I'm merging this for now as we've seen several C++/deps PRs which has gotten `lib / src` wrongfully - with these changes those PRs would not have gotten that label.

PR #55, opened 2016-07-24
labels: add version labels

mscdex:

PR #60, opened 2016-08-05
Testing webhooks and node core labels

phillipj:
This is a first shot at testing the code related to webhooks, and node core labelling specifically.

The gist of this is to POST a recorded payload against the bot's `/hooks/github` endpoint, mock responses from github.com, and verify that the bot tried to GET/POST a set of expected github.com API URLs.

While working on this I noticed a couple of unexpected URLs being requested. Found a bug in the old github npm dependency we used, so decided to update that dependency as well.

Also moved all tests into `tests/{integration,unit}` directories.

Refs #55


lib/github-events.js
```diff
@@ -16,16 +10,15 @@ module.exports = (app) => {
       return res.end()
     }
 
-    const signature = req.headers['x-hub-signature']
-    const data = req.body
-    data.action = data.action ? event + '.' + data.action : event
-
-    if (!signature || signature !== sign(secret, req.raw)) {
+    if (!githubSecret.isValid(req)) {
```
phillipj: Extracted into its own local module to make it easier to turn off the verification while running tests.

lib/node-repo.js
```diff
@@ -25,41 +25,17 @@ function updatePrWithLabels (options, labels) {
     return
   }
 
-  fetchExistingLabels(options, (err, existingLabels) => {
```
phillipj: We shouldn't need to fetch existing labels when adding PR labels, cause the updated github dependency has a new method for *adding labels*, rather than editing the whole issue like we did before.

lib/pollTravis.js
```diff
@@ -94,7 +94,7 @@ function pollTravisBuildBySha (options, checkNumber = 1) {
 
 function createGhStatusFn (options) {
   return (state, travisId, message) => {
-    githubClient.statuses.create({
+    githubClient.repos.createStatus({
```
phillipj: Changed cause of breaking changes in the github dependency

phillipj:
@nodejs/github-bot I'm done with this for now, any feedback would be much appreciated ❤️

Starefossen:
Had to read up on `knock` - but from my understanding `filesScope.done()` and `newLabelsScope.done()` will throw an exception if the requests have not been called, right?

phillipj:
> ..  filesScope.done()  and  newLabelsScope.done()  will throw an exception if the requests have not been called, right?

Yepp, which ends up breaking the tests

Starefossen:
Looks solid to me. Will `knock` throw an exception if there are other requests made which are not intercepted?

phillipj:
Hm, yeah that would throw, but I'll double check to see if those exceptions are bubbled up correctly making the tests explode

phillipj:
Found that nock returned error to the request callbacks, which didn't really make the tests explode as badly as I wanted. Just pushed a commit fixing that -- I'll probably make that `nock.emitter.on('no match', ..)` weirdness a local helper module later.

phillipj:
@Starefossen what's your preference on commits / squashing? I've tried to separate each set of changes and include a description where needed. Obviously if I squash 'em all it won't be as descriptive in the git log...

phillipj:
Decided to merge this, as it kinda stalled progress on https://github.com/nodejs/github-bot/pull/55. Kept each commit as they're pretty useful on their own, will try to keep the next PRs more focused.

Starefossen:
Sorry I dropped the ball on this one @phillipj 😢 I often like to have individual commits in the timeline if they are logical units of work with well written messages instead of squashing them.

phillipj:
> Sorry I dropped the ball on this one @phillipj 😢

No worries! Seems we agree on commits vs squashing tho :)

Issue #23, opened 2016-04-12
Bot deployment, access, & permission


williamkapke:
ATM, the bot is deployed manually by @phillipj to his dokku account while we get it started.

The bot account needs significant org permissions to perform some of it's actions. This access means PRs needs to be carefully reviewed and scrutinized for security before deployment.

We will probably want to:
- Move this repo to the nodejs org
- Give commit rights to a restricted set of people in the Node.js Org
- Determine an owner. Build WG? CTC? TSC? Something new?
- Determine a deployment strategy. (Automatic? Always Manual? By whom?)

/cc @mikeal

jbergstroem:
Since the build group runs and maintains most infra for Node.js I think this would be a good fit. Ultimately, I'd like to see an automatic deployment strategy if the active contributors in here deem it stable enough.

Fishrock123:
We could probably have this just deploy like the website, where `master` automatically gets picked up by a webhook.

Fishrock123:
@nodejs/ctc are we all comfortable with this?

Fishrock123:
> The bot account needs significant org permissions to perform some of it's actions. 

Oh true, because of write perms for labels.

I don't actually think it's the hugest deal. All the repos are git, anything you could do is pretty easily recoverable (besides deleting user comments strangely enough).

If we want it to be secure: we should make it check gpg keys for signed git tag(s).

jbergstroem:
Also likely needs elevated access to jenkins.

Fishrock123:
Hmmm, https://github.com/nodejs-github-bot/github-bot/pull/29 would need owner GitHub org access.

Perhaps it would be better to do that bit though a separate proxy or something so that we don't need to be as worried?

phillipj:
> Perhaps it would be better to do that bit though a separate proxy or something so that we don't need to be as worried?

@Fishrock123 could you elaborate on what a separate proxy means in detail/what it would do in practice, and why that would make it safer than just providing this bot with GH org access?

phillipj:
I just moved the bot repo to the org 👯

williamkapke:
All of the tasks listed have been completed! 🎉 
Approving & assigning elevated permissions should be discussed in new issues (and are).

Closing.

PR #66, opened 2016-08-25
Add docs required by TSC

williamkapke:
[As promised](https://github.com/nodejs/TSC/issues/117#issuecomment-237924423) in the [2016-08-11 TSC meeting](https://github.com/nodejs/TSC/issues/127) , here are the boilerplate docs required by the TSC for repos entering the Node.js GitHub org.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,36 @@
+# Contributing
+
+Welcome! Thanks for your interest in contributing to the Node.js GitHub Bot. If you'd like to report a 
+bug or suggest a feature, please [file an issue](https://github.com/nodejs/github-bot/issues).
+
+### Code of Conduct
+
+The [Node.js Code of Conduct][] applies to this project.
+
+[Node.js Code of Conduct]: https://github.com/nodejs/node/blob/master/CODE_OF_CONDUCT.md
+
+## Developer's Certificate of Origin 1.1
+
+By making a contribution to this project, I certify that:
+
+* (a) The contribution was created in whole or in part by me and I
+  have the right to submit it under the open source license
+  indicated in the file; or
+
+* (b) The contribution is based upon previous work that, to the best
+  of my knowledge, is covered under an appropriate open source
+  license and I have the right under that license to submit that
+  work with modifications, whether created in whole or in part
+  by me, under the same open source license (unless I am
+  permitted to submit under a different license), as indicated
+  in the file; or
+
+* (c) The contribution was provided directly to me by some other
+  person who certified (a), (b) or (c) and I have not modified
+  it.
+
+* (d) I understand and agree that this project and the contribution
+  are public and that a record of the contribution (including all
+  personal information I submit with it, including my sign-off) is
+  maintained indefinitely and may be redistributed consistent with
+  this project or the open source license(s) involved.
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -1,0 +1,5 @@
+# Node.js GitHub Bot Project Governance
+
+The Node.js GitHub Bot is overseen by the Node.js Technical Steering Committee.
+https://github.com/nodejs/TSC
+
--- a/LICENCE.md
+++ b/LICENCE.md
@@ -1,0 +1,18 @@
+The MIT License (MIT)
+=====================
+
+Copyright (c) 2016 Node.js Github Bot contributors
+--------------------------------------------------
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated 
+documentation files (the "Software"), to deal in the Software without restriction, including without limitation 
+the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and 
+to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all copies or substantial portions of 
+the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO 
+THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE 
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, 
+TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE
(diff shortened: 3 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,37 @@
+# Contributing
+
+Welcome! Thanks for your interest in contributing to the Node.js GitHub Bot. If you'd like to report a 
+bug or suggest a feature, please [file an issue](https://github.com/nodejs/github-bot/issues).
+
+### Code of Conduct
+
+The [Node.js Code of Conduct][] applies to this WG.
```
phillipj: What WG? ;)

williamkapke: Fixed!

phillipj:
Only thing which comes to mind, is what WG this project is governed by. You raised that question in https://github.com/nodejs/github-bot/issues/23, but no one has answered, and we haven't really discussed it directly in other occasions afaik.

Starefossen:
> what WG this project is governed by 

Build or Website would be candidates. The Bot WG has some resemblance/overlap with the Testing WG which is supposed[?](https://github.com/nodejs/testing/blob/master/Charter.md) to be a WG under the Build WG.

phillipj:
ATM I agree with Build, but that decision belongs to the Build overlords I assume?

In the future we might end up spawning a separate WG too handle this, e.g Automation, but that sounds like overkill for the current size of the project and amount of collabs.

/cc @jbergstroem @rvagg

jbergstroem:
I think build is a good place and would be happy to work towards that.

rvagg:
> but that decision belongs to the Build overlords

that includes you now overlord @phillipj

williamkapke:
I removed my accidental reference to a "WG".

After the bot has a deployment process, I don't find the Build group to be the appropriate group to oversee the development of the work it does.

The bot's scope is very crosscutting and will extend beyond the reach of even the CTC. This is why I pointed the Governance to the TSC.

I hope to discuss this as the [Collaboration Summit in Amsterdam](https://github.com/nodejs/summit/issues/23) and will likely continue the discussion in Austin.

Starefossen:
> The bot's scope is very crosscutting and will extend beyond the reach of even the CTC. This is why I pointed the Governance to the TSC.

While I agree with you, there are two point I would like to make. 1) The development of the bot is primarily a technical task which means that the governing body should be one of developers who have time to sit down and develop the features. 2) I see the governing body of the bot more of a service agency for the rest of the organisation - taking requests from other WGs and implementing them. This is much like how Build WG is operating today. It has the responsibility to maintain the CI and takes requests from other WGs to implement changes without necessarily being the driving force behind the change.

williamkapke:
@Starefossen I see it more like the docs WG where the process of publishing is automated by the build WG, but their content are within their own oversight.

phillipj:
> I hope to discuss this as the Collaboration Summit in Amsterdam and will likely continue the discussion in Austin.

Any reason these discussions can't happen publicly and async, like we're doing right now, rather than requiring devs to show up at summits to be part of the discussions?

jbergstroem:
I was hoping to have a hangout call prior to the summit if that works for everyone? Can suggest something in a new issue.

williamkapke:
@phillipj certainly! The summit is just an opportunity to bring awareness and solicit feedback from folks that aren't watching this progress.

phillipj:
As for Build WG vs TSC, I'd say start with Build WG and possibly elevate it to TSC when we know for a fact that it would benefit the project.

Fishrock123:
Should the license be defaulting to Apache-2.0 nowadays?

Fishrock123:
@mikeal should we be using Apache-2.0?

CONTRIBUTING.md
```diff
@@ -0,0 +1,36 @@
+# Contributing
+
+Welcome! Thanks for your interest in contributing to the Node.js GitHub Bot. If you'd like to report a 
+bug or suggest a feature, please [file an issue](https://github.com/nodejs/github-bot/issues).
+
+### Code of Conduct
+
+The [Node.js Code of Conduct][] applies to this project.
```
Starefossen: Missing link

williamkapke:
From today's TSC meeting: At the moment, MIT is the correct license according to the By-Laws. There is a desire to switch to Apache-2.0, but it will require some work.

Governance was discussed also with the conclusion that it should be listed as the TSC- but the people belonging to the GitHub team are the ones running it.

Can I get some LGTMs?

Starefossen:
Thanks for the update! LGTM

GOVERNANCE.md
```diff
@@ -0,0 +1,5 @@
+# Node.js GitHub Bot Project Governance
+
+The Node.js GitHub Bot is overseen by the Node.js Technical Steering Committee.
+https://github.com/nodejs/TSC
+
```
jbergstroem: As mentioned in the chat, should we leave this open until later? I mean, we'd still need to apply for this to be true - right?

jbergstroem:
I think we would spend some time improving the goals/rationale of this (to be) wg; this - taken from the realm - isn't necessarily true: ` All repositories that use this bot have the same webhook url & secret configured (there is only 1 bot instance). Org-wide webhooks are not allow.`

Lets try and come up with something that better reflects what we want to achieve.

GOVERNANCE.md
williamkapke: Everything in the GitHub org is under the TSC. The TSC, delegates many/most things to a dedicated group/WG/Committee.

If someday we want to create a new WG that manages the bot, we could request the TSC assigns the responsibility to the WG. But for now, it is delegated to the members of the @nodejs/github-bot team.

williamkapke:
@jbergstroem Agreed that we should update the readme. It would be nice to see "Guiding Principles" or similar.

The TSC didn't seem too keen on creating a new WG. My guess is that it is because there are a bunch of WGs that have lost their steam.

<br><br>

> All repositories that use this bot have the same webhook url & secret configured (there is only 1 bot instance).

I added that in to remove confusion for anyone. It is true at the moment- but maybe not in the future.

<br><br>

> Org-wide webhooks are not allow.

This definitely needs to hold true. Otherwise we leak from private repos.

GOVERNANCE.md
jbergstroem: Ok, cool.

phillipj:
> > Org-wide webhooks are not allow.

> This definitely needs to hold true. Otherwise we leak from private repos.

That's obviously a big issue for 3rd party GitHub apps, but wouldn't it be very different for this bot which we (the nodejs org) have full control over?

GOVERNANCE.md
phillipj: > But for now, it is delegated to the members of the @nodejs/github-bot team.

To me that sounds like we should replace TSC w/github-bot team here.

Tbh as long as the bot stays in the nodejs org I'm happy.

williamkapke: The problem with teams is that they're not viewable to people that are not members of the org. In rare scenario there is someone that needs to escalate something beyond the bot's issue tracker- we should provide them a group with an open/transparent governance body.

williamkapke:
This week's TSC meeting expressed a desire to merge this as-is and iterate as needed.

PR #70, opened 2016-09-05
labels: stricter version match in base branch

phillipj:
Minor adjustment to the regex matching the base branch a PR is targeted against.

Refs https://github.com/nodejs/github-bot/pull/55#discussion_r77553358

/cc @Fishrock123

### DIFF
--- a/lib/node-labels.js
+++ b/lib/node-labels.js
@@ -70,7 +70,7 @@ function resolveLabels (filepathsChanged, baseBranch, limitLib = true) {
                   : matchAllSubSystem(filepathsChanged, limitLib)
 
   // Add version labels if PR is made against a version branch
-  const m = /^(v\d+\.(?:\d+|x))(?:-|$)/.exec(baseBranch)
+  const m = /^(v\d+\.(?:\d+|x))(?:-staging|$)/.exec(baseBranch)
   if (m) {
     labels.push(m[1])
   }


Starefossen:
**Before:**

screen shot 2016-09-05 at 22 31 45

**After:**

screen shot 2016-09-05 at 22 31 57

mscdex:
I'm +0/leaning more towards -1 on this. What situation could arise where the current regexp would not match correctly? If anything I could see other version-specific branch suffixes, which would mean more work in maintaining the regexp. \*shrug\*

Fishrock123:
I would prefer this, since it's possible that otherwise it may match branches that are say, more specific and may create new labels unwarrented.

mscdex:
Release proposals have either been starting with `'proposal-'` or they are more specific in the version (e.g. `'v6.1.1-proposal'` whereas the current regexp would only match at most `'v6.1-'`. In fact, once v0.10/v0.12 are no longer, we can remove the second digit check altogether and at that point it should be *even less* of a problem since it will only look for '.x'.

phillipj:
Although not too opinionated about this, in the case of having more explicit code, I'm leaning towards +1 on merging this.

Regex isn't everyones favorite lunch, adding the `-staging` prefix makes it easier to grasp at first glance.

phillipj:
As far as I see, in sum the votes leaned slightly towards merging this, so here it goes to close this stalled PR.

Issue #58, opened 2016-08-05
Creating new labels?


jasnell:
When I opened https://github.com/nodejs/node/pull/7972 which suggests adding a new internal module called `mapproxy`, the bot apparently went and proactively created a new label for it. Unfortunately, `mapproxy` is not an official thing and having such a label doesn't make sense yet.

phillipj:
Good point, would be good to double check if the resolved labels actually
exists before adding them... At least for now as it has not intentionally
been made to create new labels like it just did

PR #80, opened 2016-09-19
labels: prevent new labels being created

phillipj:
As @jasnell reported in #58, the bot (unexpectedly) creates new repo labels once in a while. These changes prevents that from happening, by checking which labels already exists in the repo, before adding labels to PRs.

Closes https://github.com/nodejs/github-bot/issues/58


phillipj:
@nodejs/github-bot anyone willing to approve and share the blame on this one?

Fishrock123:
Didn't we originally do this? I have a recollection of it.

phillipj:
Earlier we fetched which labels were already added to a PR before adding
additional resolved labels, assume that's what you're referring to.

This is not related, as this ensure *new repo labels* aren't created in the
node repo. As @jasnell discovered, the bot might end up creating completely
new labels that doesn't already exist, cause a PR has created a new file in
the lib-directory, e.g lib/mapproxy.js as the PR he mentioned.

Fishrock123:
Oh right. Do we need to do this for every check, or can we do it like.. once every hour?

phillipj:
Good idea, once an hour should be sufficient. I'll fix

phillipj:
Long overdue, but I just pushed cache w/1 hour expiry for the existing repo labels as suggested. 

PTAL

phillipj:
@nodejs/github-bot this has stalled a bit, but I'll merge this within a day or two if there's no more comments or objections

Issue #79, opened 2016-09-17
Missing labels


mscdex:
I noticed that in some cases, like [this PR](https://github.com/nodejs/node/pull/8600), the bot failed to label the PR at all. It should have added the `test` label.

sejoker:
I am curious how to check log files in case we got an error from Github. In that case, we should have a message 'Error retrieving files from GitHub' in the logs.

mscdex:
Your guess is as good as mine. I've never tried to run my own instance of the bot before.

jbergstroem:
We use Bunyan so logs are available on the host as well as over http (with a secret). I'm sure @phillipj can provide the access key to those that wants to fiddle.

phillipj:
Absolutely! Just ping me on IRC and I'll give you the URL

phillipj:
Had a look in the logs and found this:

```
[2016-09-17T11:42:06.032+02:00] ERROR: bot/15719 on server-05: Error retrieving files from GitHub (req_id=8dfe80e7-e6dd-4144-9eff-5c81afb17e88, pr=nodejs/node/#8600, action=pull_request.opened)
    err: {
      "code": 404,
      "status": "Not Found",
      "message": "{\"message\":\"Not Found\",\"documentation_url\":\"https://developer.github.com/v3\"}"
    }
```
I assume there was some inconsistencies in GitHub's side which caused this error, while trying to retrieve the files related to the mentioned PR :/

Maybe the bot should wait a couple of seconds before doing its thing, as it's not important that these labels gets added *instantaneously*.

sejoker:
@phillipj thanks for follow up, you confirmed my initial assumption. A short timeout like you suggested would be sufficient as the first step. If necessary, more complicated setup with retry logic will need to take place.

sejoker:
I could prepare a PR if we agree on the possible fix.

phillipj:
@sejoker a PR would be great 👍

sejoker:
@phillipj your feedback on https://github.com/nodejs/github-bot/issues/81 will be valuable.

phillipj:
FYI now that @sejoker PR is merged, the bot will wait for 2 seconds before resolving labels

PR #81, opened 2016-09-19
labels: defer resolving labels

sejoker:
These changes defer the process of label resolving
to prevent 404 errors from Github API

Closes https://github.com/nodejs/github-bot/issues/79

### DIFF
--- a/lib/node-repo.js
+++ b/lib/node-repo.js
@@ -4,6 +4,11 @@ const githubClient = require('./github-client')
 
 const resolveLabels = require('./node-labels').resolveLabels
 
+function deferredResolveLabelsThenUpdatePr (options) {
+  const timeoutMillis = (options.timeoutInSec || 0) * 1000
+  setTimeout(resolveLabelsThenUpdatePr, timeoutMillis, options)
+}
+
 function resolveLabelsThenUpdatePr (options) {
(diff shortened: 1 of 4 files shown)


lib/node-repo.js
```diff
@@ -4,6 +4,15 @@ const githubClient = require('./github-client')
 
 const resolveLabels = require('./node-labels').resolveLabels
 
+function deferredResolveLabelsThenUpdatePr (options) {
+  if (options.timeoutInSec) {
+    setTimeout(() => resolveLabelsThenUpdatePr(options),
+              options.timeoutInSec * 1000)
+  } else {
+    resolveLabelsThenUpdatePr(options)
```
phillipj: What if we got rid of this `else` block and always executed this async? Something like:

```javascript
const timeoutMillis = (options.timeoutInSec || 0) * 1000

setTimeout(resolveLabelsThenUpdatePr, timeoutMillis, options)
```

phillipj:
Little nitpick, other than that it LGTM

lib/node-repo.js
sejoker: @phillipj updated PR as per comments, looks cleaner, thanks.

phillipj:
Awesome, thanks for contributing! 👍

Issue #86, opened 2016-10-18
Bot failed to label PR


mscdex:
https://github.com/nodejs/node/pull/9171

Not sure why it didn't label it as `stream`.

mscdex:
Another instance tonight: https://github.com/nodejs/node/pull/9173

phillipj:
I'm probably the one to blame after merging a label related PR yesterday.

Had a quick look in the logs, no errors found, and it did get the "PR opened"-event from GitHub on both PRs. Doing more verbose logging would be good for easier debugging

mscdex:
Possibly another instance FWIW: https://github.com/nodejs/node/pull/9178

mscdex:
Another: https://github.com/nodejs/node/pull/9187

Should I stop posting examples?

phillipj:
Yepp, thanks! I'll dig into this tonight, and keep you updated.

phillipj:
🎉 https://github.com/nodejs/node/pull/9191

Just found a possible corner case which haven't been fixed tho. Sadly the github client lib we're using only allows us to fetch 100 repository labels with one single call to `.getLabels()`. As the node repo currently has 117 repo labels in total, there's a chance we won't add some resolved labels as we consider those 17 labels not to exist in the repo, and therefore won't add them to PRs either.

Will have a look at improving that later.

mikeal:
Maybe we should just pair down to less that 100 labels :)

PR #87, opened 2016-10-19
labels: fix labels not being added

phillipj:
After PR https://github.com/nodejs/github-bot/pull/80 got merged, no labels were added to new PRs.

That regression probably came from the fact that the existing repository labels was fetched and compared to the label names our labelling algorithm thought we should add. The problem is that we compared the whole label meta object (from api.github.com) against a label name (string).

Example of the two different types of objects:

```js
// complete label object
{
  "url": "https://api.github.com/repos/nodejs/node/labels/buffer",
  "name": "buffer",
  "color": "f7c6c7"
}

// label name
"buffer"

```

Comparing two such objects will obviuosly never match.

These changes extracts the existing repo label *names* for comparison, and adds more logging for future debugging purposes.

Closes https://github.com/nodejs/github-bot/issues/86

/cc @mscdex

### DIFF
--- a/test/integration/node-labels-webhook.test.js
+++ b/test/integration/node-labels-webhook.test.js
@@ -1,8 +1,6 @@
 'use strict'
 
 const tap = require('tap')
-const fs = require('fs')
-const path = require('path')
 const url = require('url')
 const nock = require('nock')
 const supertest = require('supertest')
@@ -20,6 +18,8 @@ const testStubs = {
 }
 
 const app = proxyquire('../../app', testStubs)
+
+const readFixture = require('../read-fixture')
 
 setupNoRequestMatchHandler()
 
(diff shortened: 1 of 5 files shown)


test/integration/node-labels-webhook.test.js
```diff
@@ -124,11 +124,6 @@ function ignoreQueryParams (pathAndQuery) {
   return url.parse(pathAndQuery, true).pathname
 }
 
-function readFixture (fixtureName) {
```
phillipj: Extract the `readFixture()` helper method into its own submodule, as more and more tests needed it.

mscdex:
LGTM if CI is ok with it.

phillipj:
This just got deployed.

PR #89, opened 2016-10-31
Fix running individual scripts with `npm start`

phillipj:
When introducing traceable logs and bunyan in PR #42, it broke the possibility of providing individual scripts to run as second argument to `npm start`.

This fixes that by allowing scripts to be passed in as an environment variables, instead of a CLI argument to `app.js`.

Refs https://github.com/nodejs/github-bot/pull/42

/cc @Fishrock123


phillipj:
fyi @Fishrock123 

Merging this trivial change as it only affects developer experience, rather than important runtime functionality.

Issue #77, opened 2016-09-12
Report what staging branches a PR can land cleanly on


MylesBorins:
This would be VERY useful information for backporting

phillipj:
Huh, this sounds like a cool feature! Should it post a comment or add labels?

addaleax:
I think labels would be good? They are pretty cheap, and maybe it’s just me but I find automated github comments by bots more noisy than helpful.

jbergstroem:
I reckon just trying to rebase any branch against whatever branch would be pretty cheap? These doesn't have to show up in regular PR's, just one's that has a the land-on-$foo stuff?

TheAlphaNerd:
what I would personally like to see would be automating adding the
Lts-watch labels if a commit lands cleanly on that release stream

jbergstroem:
Oh - that would be pretty neat.

mscdex:
Why use the `lts-watch-*` labels? Shouldn't those be added by users whether there are merge conflicts or not?

Also, how would you know which labels should be added (which branches to test), unless you mean to try all branches that are not specifically excluded via `do-not-land-on-*` labels?

Maybe a new set of labels would be best? Like `lands-cleanly-on-*` or something?

PR #90, opened 2016-11-02
scripts: new attempt-backport script for PRs

Fishrock123:
This is a script to automatically check if PRs can cleanly land on staging branches in the node repo.

I've been using https://github.com/TestOrgPleaseIgnore/node/pull/29 and https://github.com/TestOrgPleaseIgnore/node/pull/30 to test it and it seems to be working fine now.

This is however a first iteration and I'm sure it can be improved lots.

cc @nodejs/github-bot & @TheAlphaNerd 

Edit: some unresolved questions/issues:

- Should it unlabel if the PR later passes?

### DIFF
--- a/lib/node-repo.js
+++ b/lib/node-repo.js
@@ -93,6 +93,7 @@ function itemsInCommon (arr1, arr2) {
   return arr1.filter((item) => arr2.indexOf(item) !== -1)
 }
 
+exports.updatePrWithLabels = updatePrWithLabels
 exports.resolveLabelsThenUpdatePr = deferredResolveLabelsThenUpdatePr
 
 // exposed for testability
--- a/lib/node-repo.js
+++ b/lib/node-repo.js
@@ -61,6 +61,30 @@ function updatePrWithLabels (options, labels) {
   })
 }
 
+function removeLabelFromPR (options, label) {
+  // no need to request github if we didn't resolve a label
+  if (!label) {
+    return
+  }
+
+  options.logger.debug('Trying to remove label: ' + label)
+
+  githubClient.issues.removeLabel({
+    user: options.owner,
+    repo: options.repo,
+    number: options.prId,
+    name: label
+  }, (err) => {
+    if (err) {
+      if (err.code === 404) return options.logger.info('Label to remove did not exist, bailing ' + label)
+
+      return options.logger.error(err, 'Error while removing a label')
+    }
+
+    options.logger.info('Removed a label ' + label)
+  })
+}
+
 function fetchExistingLabels (options, cb) {
   const cacheKey = `${options.owner}:${options.repo}`
 
@@ -93,6 +117,7 @@ function itemsInCommon (arr1, arr2) {
   return arr1.filter((item) => arr2.indexOf(item) !== -1)
 }
 
+exports.removeLabelFromPR = removeLabelFromPR
 exports.updatePrWithLabels = updatePrWithLabels
 exports.resolveLabelsThenUpdatePr = deferredResolveLabelsThenUpdatePr
 
--- a/scripts/attempt-backport.js
+++ b/scripts/attempt-backport.js
@@ -1,0 +1,205 @@
+'use strict'
+
+const child_process = require('child_process')
+const debug = require('debug')('attempt-backport')
+const request = require('request')
+const updatePrWithLabels = require('../lib/node-repo').updatePrWithLabels
+
+const enabledRepos = ['node']
+const queue = []
+let inProgress = false
+
+module.exports = function (app) {
+  if (!global._node_repo_dir) return
+
+  app.on('pull_request.opened', handlePrUpdate)
+  // Pull Request updates
+  app.on('pull_request.synchronize', handlePrUpdate)
+
+  function handlePrUpdate (event, owner, repo) {
+    if (!~enabledRepos.indexOf(repo)) return
+
+    const prId = event.number
+    const options = { owner, repo, prId, logger: event.logger }
+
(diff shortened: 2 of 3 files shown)


phillipj:
A little more use of the bunyan logger, *really* valuable for doing post-mortem on PRs in production. See [./lib/node-repo.js](https://github.com/nodejs/github-bot/blob/master/lib/node-repo.js) for inspiration. Other than that it looks good enough to test live IMO 👍

Couple of notes for future improvements:

- Some tests would be nice to ensure we don't break this later
- Put `queue` outside of process memory too handle redeploys

The latter is already an issue, and might be subject to a bigger discussion. E.g. for the Travis build poller ([./lib/pollTravis.js](https://github.com/nodejs/github-bot/blob/master/lib/pollTravis.js)) where a redeploy could easily make inline status on a PR stall for eternity.

Fishrock123:
> where a redeploy could easily make inline status on a PR stall for eternity.

Definitely a bug, but pushing anything to the PR branch will unblock that.

phillipj:
> Definitely a bug, but pushing anything to the PR branch will unblock that.

Definitely, tho I assume not many contributors would know that.

Fishrock123:
Updated and I tried to get it working at https://github.com/TestOrgPleaseIgnore/node/pull/31 but I wasn't able to get the new label removal working?

I have no idea how to test this, there are so many (necessary) moving parts. 😓

phillipj:
Yeah I've noticed that as well, it's generally quite painful to test much
of what the bot is doing. The test code quickly gets more complex than the
actual source code 😕

phillipj:
Should `$NODE_REPO_DIR` be an empty directory, or should the node repo already be cloned into that directory? I might get that setup in the ansible playbook before merging/deploying this.

Fishrock123:
@phillipj it will need to be a github clone.

I thought of adding cloning logic into this but decided against it.

Fishrock123:
Looping @mscdex's comment (https://github.com/nodejs/github-bot/issues/77#issuecomment-254644413) from the original thread which I somehow missed:

> Why use the `lts-watch-*` labels? Shouldn't those be added by users whether there are merge conflicts or not?

> Also, how would you know which labels should be added (which branches to test), unless you mean to try all branches that are not specifically excluded via `do-not-land-on-*` labels?

> Another issue is how would you keep the labels up-to-date? If someone hasn't updated their PR/branch in awhile, there could be merge conflicts at some point.

> If we stick with labels, maybe a new set of labels would be best? Like `lands-cleanly-on-*` or something?

> Is it possible to have a webhook like Travis and other services that shows the information at the bottom of the PR? If so, would that be better than labels?

Thoughts everyone?

TheAlphaNerd:
I think we should stick with `lts-watch`. As the primary user of them at this time what I mostly care about is getting rid of unlabelled commits. People are pretty good at knowing that a commit is not properly labelled, and every commit is going to be audited one more time before it is backported. Adding an extra label is just another step that we will have to add to the auditing process.

phillipj:
> I thought of adding cloning logic into this but decided against it.

@Fishrock123 sounds good. I'll be able to fix what's needed in the [bot's ansible playbook](https://github.com/nodejs/build/blob/master/setup/github-bot/ansible-playbook.yaml), so we're sure the server running the bot, has the repo checked out and gets `$NODE_REPO_DIR` set.

Fishrock123:
@phillipj what's your thoughts on moving this forward? Should I just try to test it more?

phillipj:
@Fishrock123 mainly getting some feedback on https://github.com/nodejs/build/issues/531, otherwise it LGTM.

After that, if you feel confident enough to test it on the actual node repo, I'm totally fine with that. Just got auto deploy upon merge to master working here, so you'll be able to deploy this yourself when you feel like it.

phillipj:
@Fishrock123 it's ready -- the bot server now has the nodejs/node repo cloned and `$NODE_REPO_DIR` set to the repo directory

phillipj:
Awesome, really looking forward to seeing this in action! 😃

Fishrock123:
https://github.com/nodejs/node/pull/9648 appeared to be a case where the script failed prematurely because I was pointing to the wrong git remote (`upstream`) as it were on my machine.

https://github.com/nodejs/github-bot/commit/187cbd866f91e66145eab800e2e1ce17919825ca should fix that by running git commands against `origin`.

mscdex:
I think having the bot applying the `dont-land-on-*` labels is a bit of a misnomer. My understanding of those labels is that they indicate *intention*, not whether the PR can currently be backported cleanly or not.

I think a different set of labels should be used by the bot, but the problem with that is that we're already hitting the 100 label ceiling (this is a separate issue that needs discussion too, because not everyone is aware of it).

Fishrock123:
> I think having the bot applying the `dont-land-on-*` labels is a bit of a misnomer. My understanding of those labels is that they indicate *intention*, not whether the PR can currently be backported cleanly or not.

We directly check these labels as part of the release process, along with `semver-*`. They already do not just indicate intention, they are used as a hard stop measure for PRs that should not land to the specified branch.

Fishrock123:
something's still not right: https://github.com/nodejs/node/pull/9472

Fishrock123:
Another: https://github.com/nodejs/node/pull/9637

https://github.com/nodejs/github-bot/commit/015096fa2813e4b23fb9b02866c6ab5d9c53a00f - disabling for now, I clearly don't have something right still, although it got further I'll need more debug output from here on.

phillipj:
Ping me if you need some help debugging on the bot server, e.g. do a clean checkout of the node repo again.

PR #98, opened 2016-12-01
fetch multiple pages of labels if necessary

not-an-aardvark:
This updates `fetchExistingLabels` to fetch *all* labels from GitHub, rather
than just the first 100 labels. This allows the bot to work correctly now
that the repo has more than 100 labels.

Refs: https://github.com/nodejs/node/issues/9841

Since `node-repo` now uses the `githubClient.hasNextPage()` function, this commit also updates the stubs in the corresponding tests to avoid throwing an error when that function is called.


phillipj:
Much appreciated, thanks for cleaning up some tests as well 👍

PR #103, opened 2016-12-13
node-repo: refactor out fetchExistingThenUpdatePr

Fishrock123:
This is a precursor to two major improvements to the attempt-backport script.

Would like to get this landed tomorrow morning so I can work on having labels only add once, and also having the bot remove it's own labels when appropriate.

### DIFF
--- a/lib/node-repo.js
+++ b/lib/node-repo.js
@@ -26,16 +26,23 @@ function resolveLabelsThenUpdatePr (options) {
     const filepathsChanged = res.map((fileMeta) => fileMeta.filename)
     const resolvedLabels = resolveLabels(filepathsChanged, options.baseBranch)
 
-    fetchExistingLabels(options, (err, existingLabels) => {
-      if (err) {
-        return options.logger.error(err, 'Error retrieving existing repo labels')
-      }
+    fetchExistingThenUpdatePr(options, resolvedLabels)
+  })
+}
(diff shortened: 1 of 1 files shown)


lib/node-repo.js
```diff
@@ -26,16 +26,23 @@ function resolveLabelsThenUpdatePr (options) {
...
+    fetchExistingThenUpdatePr(options, resolvedLabels)
+  })
+}
+
+function fetchExistingThenUpdatePr (options, labels) {
+  fetchExistingLabels(options, (err, existingLabels) => {
+    if (err) {
+      options.logger.error(err, 'Error retrieving existing repo labels')
+
+      updatePrWithLabels(options, labels)
```
phillipj: Is this intentional? Previously on [L31](https://github.com/nodejs/github-bot/pull/103/files#diff-3f74e33a82dabff73070dddacecc4dadL31) we didn't label when an error occurred

phillipj:
Other than one question, this LGTM

Fishrock123:
@phillipj Was there a reason we didn't previously label? Seems like a hole where labels could get lost if a fetch times out. Better to label twice then IMO.

phillipj:
You're right, more important to label than to being 100% sure they match
existing labels. Ignore my comment and ship it

Issue #100, opened 2016-12-08
attempt-backport still does not work Edit: now working?!


Fishrock123:
@rvagg had updated the git version to see if that was the issue was just an old git version.

I added an extra cleanup step and tested things thoroughly locally... All still worked.

https://github.com/nodejs/node/pull/10181 is an example of it failing.

It lands cleanly all the way to v4, but the bot thought it was `dont-land-on-v7.x`.

Here's some bot output:

<details>
<summary>relevant bot logs</summary>

```
12:05:27.711 DEBUG bot: processing a new backport to v7 (req_id=ac0bc597-a8ed-410f-a253-8cfdbabd7324, pr=nodejs/node/#10181, action=pull_request.opened)
12:05:27.711 DEBUG bot: aborting any previous backport attempt... (req_id=ac0bc597-a8ed-410f-a253-8cfdbabd7324, pr=nodejs/node/#10181, action=pull_request.opened)
12:05:27.716  INFO bot: request finish (req_id=ac0bc597-a8ed-410f-a253-8cfdbabd7324, duration=7, req.query={}, req.remoteAddress=::ffff:192.30.252.42, req.remotePort=61138)
    POST /hooks/github HTTP/1.1
    host: github-bot.nodejs.org:3333
    accept: */*
    user-agent: GitHub-Hookshot/4f5b94b
    x-github-event: pull_request
    x-github-delivery: 83fca580-bd68-11e6-9f8a-82fea636fda2
    content-type: application/json
    x-hub-signature: null
    content-length: 19641
    --
    HTTP/1.1 200 OK
    HTTP/1.1 200 OK
    X-Powered-By: Express
    X-Request-Id: ac0bc597-a8ed-410f-a253-8cfdbabd7324
    Date: Thu, 08 Dec 2016 17:05:27 GMT
    Connection: keep-alive
    Content-Length: 0
12:05:27.723 DEBUG bot: resetting origin/v7.x-staging... (req_id=ac0bc597-a8ed-410f-a253-8cfdbabd7324, pr=nodejs/node/#10181, action=pull_request.opened)
12:05:29.565 DEBUG bot: updating git remotes... (req_id=ac0bc597-a8ed-410f-a253-8cfdbabd7324, pr=nodejs/node/#10181, action=pull_request.opened)
[...]

PR #104, opened 2016-12-13
attempt-backport: unlabel and clean labeling

Fishrock123:
Should fix https://github.com/nodejs/github-bot/issues/100#issuecomment-266113411 while also giving a way to see when a PR update has changed the backportability.

This was always on the plans, I just hadn't gotten to it before.

cc @phillipj and maybe also @TheAlphaNerd 

I don't really trust myself that it'l work without testing though, but I have a plan on how to test the script, so I'm going to do that next and then rebase this ontop.

### DIFF
--- a/lib/node-repo.js
+++ b/lib/node-repo.js
@@ -180,7 +180,7 @@ function stringsInCommon (arr1, arr2) {
 
 exports.getBotPrLabels = getBotPrLabels
 exports.removeLabelFromPR = removeLabelFromPR
-exports.updatePrWithLabels = updatePrWithLabels
+exports.fetchExistingThenUpdatePr = fetchExistingThenUpdatePr
 exports.resolveLabelsThenUpdatePr = deferredResolveLabelsThenUpdatePr
 
 // exposed for testability
--- a/lib/node-repo.js
+++ b/lib/node-repo.js
@@ -137,6 +137,40 @@ function fetchLabelPages (options, startPageNum, cb) {
   })
 }
 
+function getBotPrLabels (options, cb) {
+  githubClient.issues.getEvents({
+    owner: options.owner,
+    repo: options.repo,
+    page: 1,
+    per_page: 100, // we probably won't hit this
+    issue_number: options.prId
+  }, (err, events) => {
+    if (err) {
+      return cb(err)
+    }
+
+    const ourLabels = []
+
+    for (const event of events) {
+      if (event.event === 'unlabeled') {
+        const index = ourLabels.indexOf(event.label.name)
+        if (index === -1) continue
+
+        ourLabels.splice(index, 1)
+      } else if (event.event === 'labeled') {
+        const index = ourLabels.indexOf(event.label.name)
+        if (index > -1) continue
+
+        if (event.actor.login === 'nodejs-github-bot') {
+          ourLabels.push(event.label.name)
+        }
+      }
+    }
+
+    cb(null, ourLabels)
+  })
+}
+
 function stringsInCommon (arr1, arr2) {
   const loweredArr2 = arr2.map((str) => str.toLowerCase())
   // we want the original string cases in arr1, therefore we don't lowercase them
@@ -144,6 +178,7 @@ function stringsInCommon (arr1, arr2) {
   return arr1.filter((str) => loweredArr2.indexOf(str.toLowerCase()) !== -1)
 }
 
+exports.getBotPrLabels = getBotPrLabels
 exports.removeLabelFromPR = removeLabelFromPR
 exports.updatePrWithLabels = updatePrWithLabels
 exports.resolveLabelsThenUpdatePr = deferredResolveLabelsThenUpdatePr
--- a/scripts/attempt-backport.js
+++ b/scripts/attempt-backport.js
@@ -5,7 +5,8 @@ const debug = require('debug')('attempt-backport')
 const request = require('request')
 const node_repo = require('../lib/node-repo')
 const updatePrWithLabels = node_repo.updatePrWithLabels
-// const removeLabelFromPR = node_repo.removeLabelFromPR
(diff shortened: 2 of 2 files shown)


lib/node-repo.js
```diff
@@ -137,15 +137,50 @@ function fetchLabelPages (options, startPageNum, cb) {
...
+
+    const ourLabels = []
+
+    for (const event of events) {
+      if (event.event === 'unlabeled') {
+        const index = ourLabels.indexOf(event.label.name)
+        if (index === -1) continue
+
+        ourLabels.splice(index, 1)
+      } else if (event.event === 'unlabeled') {
```
phillipj: Not certain this code will be reached. This the exact same `if` check as present in [line #154](https://github.com/nodejs/github-bot/pull/104/files#diff-3f74e33a82dabff73070dddacecc4dadR154) isn't it?

lib/node-repo.js
```diff
@@ -137,15 +137,50 @@ function fetchLabelPages (options, startPageNum, cb) {
...
+        const index = ourLabels.indexOf(event.label.name)
+        if (index > -1) continue
```
phillipj: `if (ourLabels.includes(event.label.name)) continue`

scripts/attempt-backport.js
```diff
@@ -129,7 +130,23 @@ function attemptBackport (options, version, isLTS, cb) {
...
+        fetchExistingThenUpdatePr(options, [`dont-land-on-v${version}.x`])
+      } else {
+        getBotPrLabels(options, (err, ourLabels) => {
+          if (err) {
+            options.logger.error(err, 'Error fetching existing bot labels')
+            return
+          }
+
+          const label = `lts-watch-v${version}.x`
+          if (ourLabels.indexOf(label) === -1) return
```
phillipj: `if (!ourLabels.includes(label)) return`

scripts/attempt-backport.js
```diff
@@ -206,12 +223,22 @@ function attemptBackport (options, version, isLTS, cb) {
...
-        // removeLabelFromPR(options, `dont-land-on-v${version}.x`)
-      // }
+        getBotPrLabels(options, (err, ourLabels) => {
+          if (err) {
+            options.logger.error(err, 'Error fetching existing bot labels')
+            return
+          }
+
+          const label = `dont-land-on-v${version}.x`
+          if (ourLabels.indexOf(label) === -1) return
```
phillipj: ditto `Array.includes()`

lib/node-repo.js
```diff
@@ -137,15 +137,50 @@ function fetchLabelPages (options, startPageNum, cb) {
   })
 }
 
+function getBotPrLabels (options, cb) {
```
phillipj: Nice 👍 This might be handy for subsystem labels as well, if we were to react on PR updates for those labels as well.

lib/node-repo.js
Fishrock123: whoops...

scripts/attempt-backport.js
```diff
@@ -129,7 +130,23 @@ function attemptBackport (options, version, isLTS, cb) {
...
+        fetchExistingThenUpdatePr(options, [`dont-land-on-v${version}.x`])
+      } else {
+        getBotPrLabels(options, (err, ourLabels) => {
+          if (err) {
+            options.logger.error(err, 'Error fetching existing bot labels')
+            return
+          }
+
+          const label = `lts-watch-v${version}.x`
+          if (ourLabels.includes(label)) return
```
phillipj: You'll need to negate this. Before I started nagging about `Array.includes()` this line was

```js
if (ourLabels.indexOf(label) === -1) return
```

that would translate to `!ourLabels.includes(label)`

scripts/attempt-backport.js
```diff
@@ -206,12 +223,22 @@ function attemptBackport (options, version, isLTS, cb) {
...
-        // removeLabelFromPR(options, `dont-land-on-v${version}.x`)
-      // }
+        getBotPrLabels(options, (err, ourLabels) => {
+          if (err) {
+            options.logger.error(err, 'Error fetching existing bot labels')
+            return
+          }
+
+          const label = `dont-land-on-v${version}.x`
+          if (ourLabels.includes(label)) return
```
phillipj: Negate this as well

phillipj:
Other than my recent negate comments, this LGTM

Fishrock123:
@phillipj On second look I think my previous checks were wrong. We only want to remove it if we added it.

phillipj:
@Fishrock123 not sure I follow you.. If we only want the bot to remove labels it added itself previously, that still means my `.includes()` comments are valid?

If I understand the current code correctly, labels added by the bot will **not** be removed since it will prevent `removeLabelFromPR()` from being reached if the label in question was actually added by the bot.

Fishrock123:
@phillipj You're right. I read it wrong. Updated.

scripts/attempt-backport.js
```diff
@@ -206,12 +223,22 @@ function attemptBackport (options, version, isLTS, cb) {
     const cp = wrapCP('git', ['am'], { stdio: 'pipe' }, function done () {
       // Success!
       if (isLTS) {
-        updatePrWithLabels(options, [`lts-watch-v${version}.x`])
-      }// else {
+        fetchExistingThenUpdatePr(options, [`lts-watch-v${version}.x`])
+      } else {
         // TODO(Fishrock123): Re-enable this, but do a check first
         // to make sure the label was set by the bot only.
```
phillipj: This comment is invalid now, right? Or were you thinking of a different kind of check?

Fishrock123: Yeah, fixed in https://github.com/nodejs/github-bot/commit/5688b42be393821fa5cf8f0e5b169cd3cd6526cf

PR #113, opened 2016-12-29
test: always run linter when running tests

mscdex:


phillipj:
Assuming the thought here is to always lint the code even though the tests fail, which sounds good to me. The downside is failing unit tests won't cause `npm test` to exit with status 1.

Example after making one test fail:

```bash
$ npm test
  ...

  63 passing (3s)
  1 failing

$ echo $?
0
```

Any thoughts on both ensuring correct exit status code and always tests + linting?

mscdex:
@phillipj Fixed.

phillipj:
Wooha, that was quick! 👍

PR #29, opened 2016-04-15
[POC] liberalize access to inviting people to the org

williamkapke:
Here is an option for @mikeal's request to [liberalize access to inviting people to the org](https://github.com/nodejs/TSC/issues/51#issuecomment-205384492).

I created a video to show it in action:
https://www.youtube.com/watch?v=3tr4yTtHQ_g
### Modify a repo's README.md

Place a special `<!-- team:name -->` comment in the repo's `readme.md`:

``` markdown
## Members
<!-- team:website - all @mentions in this section will be added to the team by the bot -->
- @williamkapke
- @Fishrock123

<!-- team -->
```

NOTE: You must specify the team's `name` in the opening comment like shown above!
### Create/update a PR

When a PR is submitted or changes are pushed, the bot will evaluate the commit(s) for changes to the `README.md` file. If changed, it will extract the content within the `<!-- team -->` comments find all `@mentions` and do a diff with the list of current team members.
### Accept the PR

Accepting a PR sends a `push` event. So, any push- even ones outside the PR process, will get picked up by the bot.

When the bot receives the `push` event, it will do the same parsing + evaluating as when the PR was created/updated, but instead of creating a comment- it applies the additions AND removals to the team (if any).

**To be very clear...**
This means anyone with commit rights can cause people to be added or removed from the team. It should also be noted that it does not validate that the repo is owned by the team being changed. In other words, someone could change the team name to cause addition/removals _to any team in the org_. Validation could be added if this is a concern.
### Assign it to the bot!

As somewhat a proof-of-concept, I added the ability to assign the PR to the bot to cause it to re-evaluate the PR. The original comment will be updated (plus an additional timestamp) with the results.

The bot will unassign itself when finished.

PR #114, opened 2017-01-01
A script to manage members of the org

williamkapke:
I have reworked the original POC code from #29 and made it ready to deploy.

/cc @nodejs/tsc 

### Modify a repo's README.md

Place a special `<!-- team:name -->` comment in the repo's `readme.md`:

``` markdown
## Members
<!-- team:website - all mentions in this section will be added to the team by the bot -->
- @williamkapke
- @Fishrock123

<!-- team -->
```

NOTE: You must specify the team's `name` in the opening comment like shown above!
### Create/update a PR

When a PR is submitted or changes are pushed, the bot will evaluate the commit(s) for changes to the `README.md` file. If changed, it will extract the content within the `<!-- team -->` comments find all `@mentions` and do a diff with the list of current team members.
### Accept the PR

Accepting a PR sends a `push` event. So, any push- even ones outside the PR process, will get picked up by the bot.

When the bot receives the `push` event, it will do the same parsing + evaluating as when the PR was created/updated, but instead of creating a comment- it applies the additions AND removals to the team (if any).

## Other changes in this PR
A few of the integration tests were changed to limit them to loading only the script it tests (via `process.env.SCRIPTS`). This prevents initialization code in unrelated scripts from running multiple times when running `npm test`.

### DIFF
--- a/scripts/manage-members.js
+++ b/scripts/manage-members.js
@@ -1,0 +1,177 @@
+'use strict'
+
+const P = require('effd')
+const request = require('request')
+const debug = require('debug')('manage-members')
+const logger = require('../lib/logger.js')
+
+var login
+const github = require('../lib/github-client.js')
+// store/cache info about the authenticated user
+github.users.get({}).then((user) => {
+  debug('authenticated as:', user.login)
+  login = user.login
+})
+.catch((err) => {
+  logger.error('authentication failed', err)
+})
+
+// we need this RegExp to have the global flag
+const mentionsRegex = new RegExp(require('mentions-regex')().source, 'g')
+const exclude = (a, b) => a.filter((x) => !b.find((y) => y === x))
+const README = (file) => /^readme\.md$/i.test(file.filename || file) // array of strings or file objects
+
+const ignoredErrors = new Set([
+  'Not Applicable',
+  'No members changed',
+  'Members section not found',
+  'Push is on non-default branch'
+])
+const logError = (err) => {
+  if (ignoredErrors.has(err && err.message)) {
+    return debug(err.message)
+  }
+  logger.error(err)
+}
+
+module.exports = function (app) {
+  app.on('push', (e, owner, repo) => onPush(e, owner, repo).catch(logError))
+  app.on('pull_request.opened', (e, owner, repo) => onPullRequest(e, owner, repo).catch(logError))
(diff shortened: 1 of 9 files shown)


williamkapke:
I'll leave this open a few more days incase someone is interested.

scripts/manage-members.js
```diff
@@ -0,0 +1,177 @@
...
+
+const P = require('effd')
+const request = require('request')
+const debug = require('debug')('manage-members')
+const logger = require('../lib/logger.js')
+
+var login
+const github = require('../lib/github-client.js')
+// store/cache info about the authenticated user
+github.users.get({}).then((user) => {
```
phillipj: I'd much rather see this be requested when it's needed, rather than run immediately at script execution. I'm suspecting this is why you need to change not-related tests by manipulating `process.env.SCRIPTS`

williamkapke: Sure- it could probably be lazily loaded. But either way- those other tests should only load the scripts they are testing.

phillipj: > But either way- those other tests should only load the scripts they are testing.

I see it from a different angle. Those tests as integration tests, with the intention of ensuring the bot as a whole plays nice together. In that sense loading all the scripts are actually valuable. E.g. a new script may add a route which overrides another script's route, meaning the pre-existing script would suddenly not be triggered at all. We won't be able to detect such issues if we only load one script at a time.

I'm all for running unit tests in complete isolation. Integration tests not so much.

scripts/manage-members.js
```diff
@@ -0,0 +1,177 @@
...
+  'Not Applicable',
+  'No members changed',
+  'Members section not found',
+  'Push is on non-default branch'
+])
+const logError = (err) => {
+  if (ignoredErrors.has(err && err.message)) {
+    return debug(err.message)
+  }
+  logger.error(err)
```
phillipj: It would be really nice if this could include more context. Rich logs has been really valuable when debugging bot issues so far.

At least a couple of things comes to mind: which repo and the related PR number

outsideris:
Any update for this?

williamkapke:
@outsideris As you saw- I've moved this effort to https://github.com/williamkapke/orgbot

Now I need to get it in to production. Thanks for you patience!

outsideris:
@williamkapke Awesome!!!!!
Where can I track the production state for node in?

PR #115, opened 2017-01-05
labels: add url-whatwg labels for WHATWG URL PRs

joyeecheung:
Ref: https://github.com/nodejs/node/issues/10625


targos:
Can you also change line 24 for the C++ files?

targos:
Maybe we should stop labeling them `url` too. /cc @jasnell

joyeecheung:
@targos I'm +1 on stop labeling them `url`. Just see what others think.

mscdex:
Yes, I would definitely say one label or the other. It makes it much more explicit. Also that way if some change/issue should affect both implementations, we can use both labels and it will be clear/make sense.

joyeecheung:
Updated to remove the `url` labels.

mscdex:
LGTM

phillipj:
Thanks! This is live 🚀

Issue #116, opened 2017-01-24
bot incorrectly adding 'dont-land-on-v7.x'


mscdex:
The bot continually adds 'dont-land-on-v7.x' even though the PR (should) lands cleanly. This has been going on for awhile now and I've had to manually remove the label in those cases.

phillipj:
Got any example PRs by any chance? I could query the logs if it's a quite recent PR.

Noticed you removed it from https://github.com/nodejs/node/pull/6601, but that seems correct cause those changes didn't land cleanly on the v7.x-staging branch when a tried on my machine a week or two ago.

mscdex:
For example: https://github.com/nodejs/node/pull/10980

or any of the recently merged test PRs (e.g. https://github.com/nodejs/node/pull/10923) which should have no problem.

MylesBorins:
tbqh I think we should consider dropping that auto label. incorrectly
labeling don't land is potentially a big problem.

mscdex:
Another example is this one: https://github.com/nodejs/node/pull/10927

This one does actually have conflicts, but not any that were not automatically resolvable (no conflicting hunks). I'm thinking the bot maybe isn't set up correctly to do these safe automatic conflict resolutions?

phillipj:
Logs related to ^ (10927):

```
06:42:51.985Z  INFO bot: Emitting GitHub event (req_id=2dabd94d-92c5-49a7-86a4-152ac4bfdc26, pr=nodejs/node/#10927, action=pull_request.synchronize)
06:42:51.986Z DEBUG bot: processing a new backport to v7 (req_id=2dabd94d-92c5-49a7-86a4-152ac4bfdc26, pr=nodejs/node/#10927, action=pull_request.synchronize)
06:42:51.986Z DEBUG bot: aborting any previous backport attempt... (req_id=2dabd94d-92c5-49a7-86a4-152ac4bfdc26, pr=nodejs/node/#10927, action=pull_request.synchronize)
06:42:52.281Z DEBUG bot: updating git remotes... (req_id=2dabd94d-92c5-49a7-86a4-152ac4bfdc26, pr=nodejs/node/#10927, action=pull_request.synchronize)
06:42:52.294Z DEBUG bot: (req_id=2dabd94d-92c5-49a7-86a4-152ac4bfdc26, pr=nodejs/node/#10927, action=pull_request.synchronize)
    Fetching origin
    
[...]

PR #117, opened 2017-01-25
attempt-backport: enable 3-way merge for git am

phillipj:
We've seen a lot of patches not applied cleanly by the bot, which in fact
should have landed without trouble, at least with 3-way merges enabled.

This trivial change enables fallback to 3-way merge when a patch
doesn't apply cleanly when running `git am`.

Should we give this a try before possibly disabling the attempt-backport labels?

Refs https://github.com/nodejs/github-bot/issues/116

/cc @mscdex @MylesBorins @Fishrock123

### DIFF
--- a/scripts/attempt-backport.js
+++ b/scripts/attempt-backport.js
@@ -214,7 +214,7 @@ function attemptBackport (options, version, isLTS, cb) {
 
   function gitAttemptBackport (req) {
     options.logger.debug(`attempting a backport to v${version}...`)
-    const cp = wrapCP('git', ['am'], { stdio: 'pipe' }, function done () {
+    const cp = wrapCP('git', ['am', '-3'], { stdio: 'pipe' }, function done () {
       // Success!
       if (isLTS) {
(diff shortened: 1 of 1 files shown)


mscdex:
It should be safe to add this, but I'm not still 100% sure that this is all that is needed for safe, automatic conflict resolution.

sam-github:
I'm pretty sure its not all that's needed.

If master has A, B, C on it, and C depends on B, then C won't cherry-pick clean, and will get marked as not applying, but when the staging branch is built from A, B, and C, it will pick clean.

This is how the check is made, right, a single commit cherry-pick/`git am`?

If it was made by cherry-picking every commit, in order, from master that hasn't been labelled by a person as don't land, that might work because its closer to how the tools are used when backporting, and supports dependencies between commits.

phillipj:
Okey, so it's probably not going to be rock solid after these changes either. But we agree that falling back to 3-way merges won't do any harm or be even more incorrect than the bot is currently, right?

sam-github:
Yes

PR #119, opened 2017-01-29
logs: configurable logs directory with $LOGS_DIR

phillipj:
The main objective behind these changes are to not write logs into the project's `./logs` directory, cause that ends up being deleted whenever the bot is re-deployed. Having some days of historic logs are very valuable while debugging any issue.

Requires deployment changes: https://github.com/nodejs/build/pull/602

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -7,20 +7,26 @@ const isRunningTests = process.env.npm_lifecycle_event === 'test'
 const stdoutLevel = isRunningTests ? 'FATAL' : 'INFO'
 
 const daysToKeepLogs = process.env.KEEP_LOGS || 10
+const logsDir = process.env.LOGS_DIR || ''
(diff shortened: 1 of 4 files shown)


phillipj:
Now that nodejs/build#602 has landed, this is also ready to be merged -- if someone dares to approve it.

lib/logger.js
```diff
@@ -7,20 +7,26 @@ const isRunningTests = process.env.npm_lifecycle_event === 'test'
 const stdoutLevel = isRunningTests ? 'FATAL' : 'INFO'
 
 const daysToKeepLogs = process.env.KEEP_LOGS || 10
+const logsDir = process.env.LOGS_DIR || ''
```
jbergstroem: in `app.js` you default to `undefined`. Perhaps choose one for consistency.

phillipj: Sounds like a good idea 👍

Issue #82, opened 2016-09-30
Automatic CI kickoff


jasnell:
It would be fantastic if the bot could automatically kick off a new CI run based on a request in comments. For instance, if I posted a comment like: `@nodejs-github-bot run CI please`.

addaleax:
How about letting the bot do that when there has not been a previous CI run and the first “approved” review is made? That seems like a reasonably automated way of implementing this to me, but I can’t really tell how tricky that would be.

phillipj:
That's not too difficult, as far as I know Jenkins has a decent API for triggering builds. My only concern is the reasoning behind the `CERTIFY_SAFE` option we have to manually check when starting a PR build:

<img width="907" alt="screen shot 2017-01-30 at 21 08 19" src="https://cloud.githubusercontent.com/assets/1231635/22439543/5384fbe0-e730-11e6-8b19-6f3f433516d6.png">

Maybe as long as the approval comes from a collaborator, that is as good as checking the `CERTIFY_SAFE` option?

addaleax:
> Maybe as long as the approval comes from a collaborator, that is as good as checking the `CERTIFY_SAFE` option?

Yeah, that’s what I had in mind. Sorry, forgot for a minute that non-collaborators can approve, too ;)

jbergstroem:
The logical next step would be just starting a job if a committer files a pr, no?

gdams:
could we not just go down the route of using https://wiki.jenkins-ci.org/display/JENKINS/GitHub+pull+request+builder+plugin though?

gdams:
could we not just go down the route of using https://wiki.jenkins-ci.org/display/JENKINS/GitHub+pull+request+builder+plugin though?

gibfahn:
This is something we're looking at for citgm as well. I think what's been suggested so far makes sense, triggering a build when:
 - The person who raised the PR has commit access to the project
 - Someone who has commit access approves the PR.

I think it would probably be worth having a manual trigger again (e.g. for rerunning CI). Maybe something like `github-bot run CI` as a comment from someone with commit access?

jbergstroem:
@gibfahn ..or we could rebuild on any pushes to the branch [again, making the assumption that a committer filed the PR]?

gibfahn:
@jbergstroem we could definitely do that (I guess as well as the above options rather than instead of?), the only question is whether we'd end up doing too many CI runs for branches that are updated frequently.

[...]

PR #123, opened 2017-02-04
jenkins: allow citgm builds to push Jenkins build updates

phillipj:
This changes allows [citgm](https://github.com/nodejs/citgm) builds to push updates like we already do for nodejs/node. That is done by making the Jenkins status update endpoints more generic, resolving the repository name from an URL parameter.

In practise this means the endpoints stay the same, but configurable:
- `/node/jenkins/start`
- `/citgm/jenkins/start`

Repository names are validated against a whitelist.

Refs https://github.com/nodejs/github-bot/issues/82#issuecomment-276715266

/cc @jbergstroem @gdams

### DIFF
--- a/scripts/jenkins-status.js
+++ b/scripts/jenkins-status.js
@@ -1,34 +1,45 @@
 'use strict'
 
 const pushJenkinsUpdate = require('../lib/push-jenkins-update')
+const enabledRepos = ['citgm', 'node']
 
 module.exports = function (app) {
-  app.post('/node/jenkins/start', (req, res) => {
+  app.post('/:repo/jenkins/start', (req, res) => {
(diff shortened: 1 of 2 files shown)


Fishrock123:
I'm a bit confused... Does this allow CITGM runs on a PR to be reported back to nodejs/node? Or just to nodejs/citgm?

phillipj:
No cross communication. Nothing will change for nodejs/node, we just enable citgm builds to report back to nodejs/citgm

gdams:
Is this ready to use now? or do we need to make some changes to the citgm repo?

phillipj:
I think it is ready to be tested. @jbergstroem has some scripts curl-ing the bot on these endpoints whenever builds starts and ends.

Issue #88, opened 2016-10-21
Bot failed to label PR


mscdex:
https://github.com/nodejs/node/pull/9221

phillipj:
Logs says it the bot never got a `pull_request.opened` event from github for that PR, only `pull_request.edited` which is strange. Never seen this happen before, might be good to take notice when this occurs and maybe ask github support if we see more of this.

mscdex:
Saw another instance just a bit ago: https://github.com/nodejs/node/pull/9332

phillipj:
#9332 seems to be related to the fact that we only fetch the first 100 labels of the node repository.

Pulled from the logs:
```
07:27:13.409 DEBUG bot: Filled existing repo labels cache: confirmed-bug,duplicate,help wanted,invalid,question,wontfix,in progress,tls,crypto,buffer,http,https,assert,util,cluster,child_process,dgram,c++,doc,debugger,dns,domain,events,fs,module,net,os,querystring,readline,repl,stream,timers,url,vm,zlib,windows,install,semver-minor,semver-major,discuss,build,V8,test,freebsd,path,meta,string_decoder,benchmark,openssl,npm,land-on-master,good first contribution,feature request,process,console,libuv,mentor-available,os x,arm,tools,land-on-v3.x,tty,intl,memory,land-on-v0.12,land-on-v0.10,land-on-v4.x,post-mortem,lts-landed-on-v4.x,lts-watch-v4.x,smartos,lts,ctc-agenda,ctc-meeting,v0.10,v0.12,v4.x,master,v5.x,lts-agenda,lib / src,lts-watch-v0.10,lts-watch-v0.12,land-on-v5.x,unconfirmed,stalled,mips,dont-land-on-v5.x,security,dont-land-on-v4.x,performance,error-handling,repro-exists,async_wrap,known issue test,promises,docs-requested,blocked,dont-land-on-v6.x,v6.x (req_id=d5d46425-aff7-411c-892b-bbacb61bec85, pr=nodejs/node/#9332, action=pull_request.opened)
```

`cares` is not part of the cached list of labels :/

mscdex:
Ah ok

mscdex:
https://github.com/nodejs/node/pull/9422 was not automatically tagged as V8. The number of labels is at 100 exactly, so that shouldn't have been a problem?

phillipj:
You're right, the `V8` label existed in the labels cache at that moment, so that's not the issue for that PR. Related logs:

```
[...]

PR #130, opened 2017-03-22
labels: retry request for PR files if GitHub request fails

phillipj:
Trying to get around some intermittent GitHub API 404 failures, probably due to a race condition where the API hasn't yet got to update its PR cache or something similar.

Refs https://github.com/nodejs/github-bot/issues/88#issuecomment-288319518

### DIFF
--- a/lib/node-repo.js
+++ b/lib/node-repo.js
@@ -1,10 +1,13 @@
 'use strict'
 
 const LRU = require('lru-cache')
+const retry = require('async').retry
 
 const githubClient = require('./github-client')
 const resolveLabels = require('./node-labels').resolveLabels
 const existingLabelsCache = new LRU({ max: 1, maxAge: 1000 * 60 * 60 })
+
+const fiveSeconds = 5 * 1000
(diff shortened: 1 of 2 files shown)


mscdex:
IMHO we shouldn't pull in `async` for something that could be accomplished with `setTimeout()` and a counter.

phillipj:
See your point in `async` being overkill for *one* function. Though writing something similar ourself should ideally include tests for that particular code, as well as maintaining it over time. Since this is strictly backend and not frontend where we'd consider JS-bundle sizes, I personally see the benefits greater than the downsides in including a battle tested package like `async`.

Got any thoughts on where the threshold should be for knowing when to use a 3rd party package or write it ourself?

*P.S. I really had two `async` use cases in mind when open this PR -> #128 / https://github.com/nodejs/github-bot/pull/128/commits/458fdc6505217d2deef3871df74d857554db7225#diff-3b3f42a9fe8032799be4775b56e049dfR14 as well.*

PR #138, opened 2017-04-15
Change "V8" labelling to "V8 Engine"

gibfahn:
Refs: https://github.com/nodejs/CTC/issues/103

Not sure if this is all the places it should be changed, I did an `rg V8` and changed the places that seemed to make sense, but review would be much appreciated.

cc/ @addaleax


phillipj:
@gibfahn nice 👍 

I found more occurrences of `v8`, so I took the liberty of pushing a new commit to your branch --
 PTAL.

gibfahn:
@phillipj whoops, thanks for that!

Feel free to land if you think it's good to go.

PR #140, opened 2017-05-01
labels: add 'postmortem' to v8 postmortem script

bnoordhuis:


lib/node-labels.js
```diff
@@ -70,6 +70,7 @@ const subSystemLabelsMap = new Map([
   // libuv needs an explicit mapping, as the ordinary /deps/ mapping below would
   // end up as libuv changes labeled with "uv" (which is a non-existing label)
   [/^deps\/uv\//, 'libuv'],
+  [/^deps\/v8\/tools\/gen-postmortem-metadata\.py/, ['V8 Engine', 'postmortem']],
```
addaleax: Fyi, the label as we currently have it is called `post-mortem`, with a dash

bnoordhuis: I'd swear it didn't have a dash this morning...  updated at any rate.

Issue #61, opened 2016-08-08
Whitelist of ip's for incoming http posts


jbergstroem:
We should be able to limit from where we accept incoming http posts so we don't have to rely on obscurity for endpoint url.

williamkapke:
I would love to say this is a MUST have- however, with a grain of salt.

Someone can just setup an org and point their webhooks at the bot and get around it. This means it will be extremely important for scripts to verify the source org in the payload... well, really, this should probably be done at the core of the bot rather than in each script. It's easy to add.

I'm not sure if Github guarantees the ip address(s?) will not change.

Webhooks allow you to set a "secret"- we should utilize this too.

jbergstroem:
For github its more appropriate to use a secret as part of the uri; but since we can control Jenkins<>gh bot we should use a whitelist for that.

phillipj:
@jbergstroem how do we get a hold of the IPs to whitelist? Copy-n-paste from [build/setup](https://github.com/nodejs/build/tree/master/setup)? Or maybe request the master Jenkins instance to validate there's a slave with a given IP?

jbergstroem:
@phillipj it will be a small list, almost never changing. Just adding an environment variable where we can throw ip's separated by space/comma would be fine.

Starefossen:
Whitelist will be applied to unauthenticated requests without a token, right?

jbergstroem:
> @Starefossen said:
> Whitelist will be applied to unauthenticated requests without a token (aka Jenkins), right?

Yes, but we could use both as well. No point in allowing anyone but Jenkins hosts to talk to that api anyway.

phillipj:
@jbergstroem now that this is merged.. Any tips on getting a hold of the IPs of all the Jenkins workers?

PR #142, opened 2017-05-06
jenkins: whitelist IPs allowed to push status changes

phillipj:
This is needed to ensure not everyone on the internet can push an inline status to any PR if they know the bot URL.

@jbergstroem does this seem sufficient?

Closes https://github.com/nodejs/github-bot/issues/61

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,8 @@ See [CONTRIBUTING.md](CONTRIBUTING.md).
   The webhook secret that GitHub signs the POSTed payloads with. This is created when the webhook is defined. The default is `hush-hush`.
 - **`TRAVIS_CI_TOKEN`**<br>
   For scripts that communicate with Travis CI. Your Travis token is visible on [yourprofile](https://travis-ci.org/profile) page, by clicking the "show token" link. Also See: https://blog.travis-ci.com/2013-01-28-token-token-token
+- **`JENKINS_WORKER_IPS`**<br>
+  List of valid Jenkins worker IPs allowed to push PR status updates, split by comma: `192.168.1.100,192.168.1.101`.
 - **`JENKINS_API_CREDENTIALS`** (optional)<br>
   For scripts that communicate with Jenkins on http://ci.nodejs.org. The Jenkins API token is visible on
   your own profile page `https://ci.nodejs.org/user/<YOUR_GITHUB_USERNAME>/configure`, by clicking the
--- a/scripts/jenkins-status.js
+++ b/scripts/jenkins-status.js
@@ -3,6 +3,21 @@
 const pushJenkinsUpdate = require('../lib/push-jenkins-update')
 const enabledRepos = ['citgm', 'node']
 
+const jenkinsIpWhitelist = process.env.JENKINS_WORKER_IPS ? process.env.JENKINS_WORKER_IPS.split(',') : []
+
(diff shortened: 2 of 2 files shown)


scripts/jenkins-status.js
```diff
@@ -3,6 +3,21 @@
 const pushJenkinsUpdate = require('../lib/push-jenkins-update')
 const enabledRepos = ['citgm', 'node']
 
+const jenkinsIpWhiteliste = process.env.JENKINS_SLAVE_IPS ? process.env.JENKINS_SLAVE_IPS.split(',') : []
```
addaleax: Is jenkinsIpWhitelist**e** a typo?

README.md
```diff
@@ -18,6 +18,8 @@ See [CONTRIBUTING.md](CONTRIBUTING.md).
   The webhook secret that GitHub signs the POSTed payloads with. This is created when the webhook is defined. The default is `hush-hush`.
 - **`TRAVIS_CI_TOKEN`**<br>
   For scripts that communicate with Travis CI. Your Travis token is visible on [yourprofile](https://travis-ci.org/profile) page, by clicking the "show token" link. Also See: https://blog.travis-ci.com/2013-01-28-token-token-token
+- **`JENKINS_SLAVE_IPS`**<br>
```
addaleax: Is there any technical reason to use that specific word?

phillipj: Slave? I thought that was the term one used for a server that executes Jenkins builds? I'm all ears to other alternatives

scripts/jenkins-status.js
phillipj: For sure, thx 👍

README.md
gibfahn: I think we use `host` and `worker` in build to avoid the issue (so `worker` makes more sense here anyway).

https://github.com/nodejs/build/tree/master/ansible#getting-things-done

phillipj:
Fixed commented issues and force pushed.

scripts/jenkins-status.js
```diff
@@ -3,6 +3,21 @@
 const pushJenkinsUpdate = require('../lib/push-jenkins-update')
 const enabledRepos = ['citgm', 'node']
 
+const jenkinsIpWhitelist = process.env.JENKINS_WORKER_IPS ? process.env.JENKINS_WORKER_IPS.split(',') : []
+
+function isJenkinsIpWhitelisted (req) {
+  const ip = req.headers['x-forwarded-for'] || req.connection.remoteAddress
```
addaleax: Is the `x-forwarded-for` header used in practice? Any client can set it some arbitrary value, so I am not sure this is foolproof

phillipj: Good question. I'm not totally confident it's actually needed no, would depend if there's some proxy in between the workers and the bot I assume.

Very good point that allows for side stepping the whole whitelist :/

If @jbergstroem don't have other thoughts on this, I'm okey with giving it a shot without using `x-forwarded-for`.

phillipj: As no objections has been made, I just pushed a commit removing the use of `x-forwarded-for` -- thanks for pointing out that security flaw.

phillipj:
Rebased and force pushed to fix merge conflict with recent updates to master.

PR #158, opened 2017-11-16
travis: improve logging, handle more states, and make the retries configurable

joyeecheung:
This helped me when debugging the setup of node-core-utils. Also I think the number of retries is too high, 20 (10mins) should already be enough.


joyeecheung:
Hmm, just looked into https://travis-ci.org/nodejs/llnode/jobs/303007637 , and I think 30min is probably a more suitable default.

joyeecheung:
Just looked into https://github.com/nodejs/llnode/pull/148 a bit, turns out we don't handle the `errored` and `canceled` states so if a Travis build times out (stalled), the bot won't be able to update the status properly. I've added handling for these two.

phillipj:
IIRC the many retries was set after seeing how long some of the docker-node builds run. For all other projects, that many retries is way too many. If that's still the case for docker-node I don't know.

joyeecheung:
@phillipj In that case we can probably make a per-repo map for retries and go over their recent Travis builds for a reasonable value? And fallback to a smaller default value for most repos

joyeecheung:
@phillipj OK, I've made the configuration a map and added tests for both `pollTravis.js` and `display-travis-status.js`, PTAL

phillipj:
You've got enough access to merge this yourself, right? ..it'll get deployed to production immediately 🚀

joyeecheung:
@phillipj Merged, thanks!

PR #161, opened 2017-11-27
jenkins: handle IPv4 remote address converted to IPv6

phillipj:
After the Jenkins IP whitelist was activated, the inline PR status reports for nodejs/node seems to be broken.

Seems like the issue here is that IPv4 remote addresses are converted to IPv6, which wasn't expected too happen.. Stripping any IPv6 parts off the remote addresses should make our IPv4 comparisons work as expected.

Refs https://github.com/nodejs/build/issues/1016
Refs https://github.com/nodejs/build/pull/985

/cc @maclover7


phillipj:
It's live

PR #174, opened 2018-06-02
jenkins: fix small bugs in trigger-jenkins-build

maclover7:


maclover7:
Getting this out now to stop the bot from continually crashing...

phillipj:
Good catch!

I assume an unhandled exception handler that logged the error would have given us some hints about what caused the unexpected restarts?

PR #184, opened 2018-06-13
jenkins: only trigger on nodejs/node PRs

maclover7:


phillipj:
To add some context for future ref, this ensures we only trigger automatic Jenkins builds for new PRs in nodejs/node, as long as the PR author is a collaborator.

maclover7:
Thanks for the quick reviews, landed in 3d60243

PR #205, opened 2018-11-24
remove redundant indent_style from editorconfig

thefourtheye:
`indent_style` is applied to all the files based on `[*]` entry. We
don't have to specify the same for `package.json` explicitly.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -16,5 +16,4 @@ indent_size = 2
 trim_trailing_whitespace = false
 
 [package.json]
-indent_style = space
 indent_size = 2


addaleax:
Potentially a dumb question, but: The values in the `[*]` and the `[package.json]` sections are different – should we adjust the indentation in the file somehow along with this?

thefourtheye:
> should we adjust the indentation in the file somehow along with this?

You mean the `package.json` file? That already uses two space indentation only. Or did you mean adjusting the indentation for rest of the files in the repo to four as in the `[*]` section?

thefourtheye:
> Oops, mixed up indent_style vs indent_size options.

😅 No problem. Thanks for the review 🙂

PR #209, opened 2019-01-15
scripts: fix reading credentials for jenkins

thefourtheye:
The existing code assumes that the credentials are already in base64,
but they are actually not. The code has to read the credentials in plain
text and base 64 encode them.

Ref: https://github.com/nodejs/github-bot/pull/206#discussion_r236054041

---

@maclover7 @phillipj

### DIFF
--- a/scripts/trigger-jenkins-build.js
+++ b/scripts/trigger-jenkins-build.js
@@ -52,7 +52,7 @@ function buildParametersForRepo (options, repo) {
 
 function triggerBuild (options, cb) {
   const { repo } = options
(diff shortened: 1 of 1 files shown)


maclover7:
https://github.com/nodejs/github-bot/pull/206#discussion_r236054041 😬 😬

phillipj:
Ohmy, not sure why I didn't see that suggested change while reviewing & landing last week.

Since this was originally suggested by @maclover7 and @thefourtheye opened this PR, I assume agree this should be headed to production :)

thefourtheye:
@maclover7 Sorry, I couldn't address your comment on time. That is why I had to open a new PR with your suggested changes.

PR #214, opened 2019-02-26
revise failure text

Trott:
* Correct spelling of _occurred_.
* Capitalize first word.
* Add punctuation.

### DIFF
--- a/scripts/trigger-jenkins-build.js
+++ b/scripts/trigger-jenkins-build.js
@@ -109,7 +109,7 @@ function triggerBuildIfValid (options) {
     triggerBuild(options, function onBuildTriggered (err, buildUrl) {
       if (err) {
         logger.error(err, 'Error while triggering Jenkins build')
-        return createPrComment(options, `@${options.author} sadly an error occured when I tried to trigger a build :(`)
+        return createPrComment(options, `@${options.author} Sadly, an error occurred when I tried to trigger a build. :(`)
       }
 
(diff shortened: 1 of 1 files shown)


scripts/trigger-jenkins-build.js
```diff
@@ -109,7 +109,7 @@ function triggerBuildIfValid (options) {
     triggerBuild(options, function onBuildTriggered (err, buildUrl) {
       if (err) {
         logger.error(err, 'Error while triggering Jenkins build')
-        return createPrComment(options, `@${options.author} sadly an error occured when I tried to trigger a build :(`)
+        return createPrComment(options, `@${options.author} Sadly, an error occurred when I tried to trigger a build. :(`)
```
refack: nit: `sadly` is not the first word in the sentence, the hail is.

Trott: I think it's debatable whether that is part of the sentence or simply something put there to notify the right person/get their attention. I wouldn't write: "Refael sadly an error occurred." It would be "Refael, sadly an error occurred." Or "Refael: Sadly an error occurred." The lack of punctuation after the hail makes it seem to me more "correct" to say that it is *not* part of the sentence. But that's certainly only my opinion and others may surely disagree. 
¯\\_(ツ)_/¯

If I had my druthers, it would probably be a colon:  `@example: Sadly, an error occurred.` If that seems good to you, I'd be happy to do that.

PR #221, opened 2019-03-20
Cache npm dependencies in Travis CI

phillipj:
While looking for some low hanging chores to do, I stumbled upon `.travis.yml`. Thought we might as well tell Travis to cache the npm dependencies to speed up our builds.

Knowing we've got the entire dependency tree locked down with `package-lock.json`, it can't hurt to cache them, right?

Refs https://docs.travis-ci.com/user/caching/#npm-cache

/cc @nodejs/github-bot

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
 language: node_js
 node_js:
   - 8
+cache: npm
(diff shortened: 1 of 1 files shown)


phillipj:
Thanks for lightning quick reviews! 💯

Issue #212, opened 2019-02-09
Post the CI link to the PR when a CI is started


joyeecheung:
Since human could forget to do this.

From a glance this should probably be done around https://github.com/nodejs/github-bot/blob/a76688e33fc1e274f3d8df35a0571750d0f11b7a/scripts/jenkins-status.js#L45-L49

phillipj:
I'm all for automating all the things!

Think you're right about which lines to hook into.. We just gotta ensure we don't post comments about *all* jobs, called "checks" in GitHub UI (see below). I assume the job we should be filtering on is `node-test-pull-request`, right?

<img width="774" alt="Screen Shot 2019-03-20 at 21 51 06" src="https://user-images.githubusercontent.com/1231635/54718303-7e52b480-4b5a-11e9-9399-39e45f46fb03.png">

joyeecheung:
@phillipj Yes, at least would could start from there.

PR #222, opened 2019-03-21
Improve integration tests for pushing Jenkins statuses

phillipj:
*This is a precursor to fixing #212.*

These changes are primarily done so we can get more reliable integration tests. Previously the incoming requests was responded to, before we knew whether or not the Jenkins status has been successfully pushed to GitHub.

That makes it challenging to know when we can assert what we want to verify in integration tests, as we don't know when the entire operation has finished.

Although the previous approach worked in practise when run in production, it should be just as important to have reliable tests that we can rely on when doing changes going forward.

/cc @nodejs/github-bot

### DIFF
--- a/test/integration/push-jenkins-update.test.js
+++ b/test/integration/push-jenkins-update.test.js
@@ -19,7 +19,10 @@ tap.test('Sends POST requests to https://api.github.com/repos/nodejs/node/status
                   .reply(201)
 
   t.plan(1)
-  t.tearDown(() => prCommitsScope.done() && scope.done())
+  t.tearDown(() => {
+    prCommitsScope.done()
+    scope.done()
+  })
 
   supertest(app)
     .post('/node/jenkins/start')
@@ -40,7 +43,10 @@ tap.test('Allows repository name to be provided with URL parameter when pushing 
                   .reply(201)
 
   t.plan(1)
-  t.tearDown(() => prCommitsScope.done() && scope.done())
+  t.tearDown(() => {
+    prCommitsScope.done()
+    scope.done()
+  })
 
   supertest(app)
     .post('/citgm/jenkins/start')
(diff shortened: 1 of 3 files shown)


test/integration/push-jenkins-update.test.js
```diff
@@ -19,7 +19,10 @@ tap.test('Sends POST requests to https://api.github.com/repos/nodejs/node/status
                   .reply(201)
 
   t.plan(1)
-  t.tearDown(() => prCommitsScope.done() && scope.done())
```
phillipj: The previous approach of testing if network requests had been sent from the bot, via nock scopes, didn't test more then one network request.

That's because we tested all scoped in a one-liner like the above.

It turns out that the above will not invoke the second `scope.done()` (because `.done() returns `undefined`), meaning only the first nock scope (read: outgoing network request) was asserted. By going away from that one-liner and invoking one by one on each line, tests starts to explode.

refack: Why do we have to wait for the `tearDown`?

phillipj: That's a great question!

I don't have a good answer to why it was placed there in the first place to be honest. Seeing it now, it surely would be more intuitive put alongside the bot's response assertion instead, something like;

```diff
-  t.tearDown(() => prCommitsScope.done() && scope.done())

  supertest(app)
    .post('/node/jenkins/start')
    .send(jenkinsPayload)
    .expect(201)
    .end((err, res) => {
+      prCommitsScope.done()
+      scope.done()
       t.equal(err, null)
    })
```

Is that along the lines of what you had in mind too @refack?

refack: Yeah, that _looks_ more "idiomatic" BDD ¯\\_(ツ)_/¯

PR #223, opened 2019-03-23
labels: add "worker" label

ZYSzys:
The PRs affected `worker` module need to be labeled manually at present.

Refs: https://github.com/nodejs/node/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc+label%3Aworker

### DIFF
--- a/test/unit/node-labels.test.js
+++ b/test/unit/node-labels.test.js
@@ -665,7 +665,8 @@ tap.test('label: "build" when ./android-configure has been changed', (t) => {
 
 [
   [ ['worker'],
-    ['lib/internal/worker.js',
+    ['lib/worker_threads.js',
+     'lib/internal/worker.js',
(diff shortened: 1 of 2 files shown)


test/unit/node-labels.test.js
```diff
@@ -658,6 +658,28 @@ tap.test('label: "build" when ./android-configure has been changed', (t) => {
 
       t.same(resolved, labels)
 
+      t.end()
+    })
+  }
+});
+
+[
+  [ ['worker'],
```
refack: Does this catch changes to [`lib/worker_threads.js`](https://github.com/nodejs/node/blob/master/lib/worker_threads.js)?

ZYSzys: Added now.

I forgot this before😅.

PR #224, opened 2019-03-23
Post CI comment in PR for node-test-pull-request builds

phillipj:
These changes posts a CI style comment to PR's whenever Jenkins reports that node-test-pull-request builds has started.

Closes https://github.com/nodejs/github-bot/issues/212

/cc @nodejs/github-bot 

*P.S. temporary non-master base branch as these changes builds on https://github.com/nodejs/github-bot/pull/222.*

### DIFF
--- a/lib/push-jenkins-update.js
+++ b/lib/push-jenkins-update.js
@@ -14,7 +14,7 @@ function pushStarted (options, build, cb) {
 
   const optsWithPr = Object.assign({ pr }, options)
 
-  if (build.identifier === 'node-test-pull-request') {
+  if (build.identifier === 'node-test-pull-request' && build.status === 'pending') {
     createPrComment(Object.assign({ number: pr }, options), `CI: ${build.url}`)
   }
 
--- a/lib/push-jenkins-update.js
+++ b/lib/push-jenkins-update.js
@@ -3,6 +3,7 @@
 const url = require('url')
 
 const githubClient = require('./github-client')
+const { createPrComment } = require('./github-comment')
 
 function pushStarted (options, build, cb) {
   const pr = findPrInRef(build.ref)
@@ -12,6 +13,10 @@ function pushStarted (options, build, cb) {
   const logger = options.logger.child(traceFields, true)
 
   const optsWithPr = Object.assign({ pr }, options)
+
+  if (build.identifier === 'node-test-pull-request') {
+    createPrComment(Object.assign({ number: pr }, options), `CI: ${build.url}`)
+  }
 
   findLatestCommitInPr(optsWithPr, (err, latestCommit) => {
     if (err) {
--- a/lib/github-comment.js
+++ b/lib/github-comment.js
@@ -1,0 +1,16 @@
+'use strict'
+
+const githubClient = require('./github-client')
+
+exports.createPrComment = function createPrComment ({ owner, repo, number, logger }, body) {
+  githubClient.issues.createComment({
+    owner,
+    repo,
+    number,
+    body
+  }, (err) => {
+    if (err) {
+      logger.error(err, 'Error while creating comment on GitHub')
+    }
+  })
+}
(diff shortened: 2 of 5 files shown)


lib/push-jenkins-update.js
```diff
@@ -13,6 +14,10 @@ function pushStarted (options, build, cb) {
 
   const optsWithPr = Object.assign({ pr }, options)
 
+  if (build.identifier === 'node-test-pull-request') {
+    createPrComment(Object.assign({ number: pr }, options), `CI: ${build.url}`)
```
refack:
```suggestion
	let message = ''
	switch(build.status) {
		case 'pending':
		    message = `CI: ${build.url}`)
		case 'failure':
			message = 'CI job failed.'
		case 'success':
		case 'unstable':
			message = 'CI job passed.'
		default:
			// ¯\\_(ツ)_/¯
			return;
	}
    createPrComment(Object.assign({ number: pr }, options), message)
```

We get at least two messages for each job.
`'pending'` when the job starts. And then the final status.

refack: FYI I wrote this via GitHub so 90% chance this has syntax error, consider this pseudo-code.

phillipj: Huh, good suggestion.

Currently the bot handles [starting](https://github.com/nodejs/github-bot/blob/master/scripts/jenkins-status.js#L25) and [ending](https://github.com/nodejs/github-bot/blob/master/scripts/jenkins-status.js#L54) of jobs separately. The changes proposed in this PR only affects jobs that has just _started_, not _ended_ like the failures/success/unstable statuses.

Previously I've gotten the impression there are concerns from collaborators about the bot "spamming" PRs with comments. Is that something to consider before introducing what you suggest here, or have collaborators been wanting this for a while already?

Btw reducing the duplication seen handling starting & ended jobs surely sounds like a good fit for a later PR though.

refack: __Personally__ I'd love to have the "ending" message... It will trigger a notification which is something that will help my workflow. a la https://github.com/nodejs/build/issues/1502.

I know it's probably written somewhere that we should track when we start a CI job, but GitHub already does that via the "status", so IDK.

/Ping @Trott 

Anyway I'd think at minimum we should:

```suggestion
    if (build.identifier === 'node-test-pull-request' && build.status === 'pending) {
```

phillipj: Adding that `build.status === 'pending'` sounds good for now IMO. It'll also serve as an explicit reminder if the handling of starting & ending jobs gets refactored into one later.

phillipj:
Rebased, force pushed and fixed pending status check after https://github.com/nodejs/github-bot/pull/222 got merged (which caused some merge conflicts).

refack:
image
🎉

phillipj:
Yeay very cool!

refack:
image

So it posts for "Resume" as well...
We might need an opt-out mechanism (less harsh them a GitHub "block user")

phillipj:
Could you elaborate what you'd rather see in the example you posted above? I'm not  entirely sure what you mean by "resume", but since all those URLs are different, is it wrong to treat them like different builds all together?

First thought that came to mind to opt-out is allowing collaborators to add a special kind of label to the PR, e.g. `bot-no-ci-comment` or similar 🤷‍♂️

refack:
Just to be clear, I'd wait until someone actually complains.

> what you'd rather see

I'm perfectly happy with the status quo, but others might find this inconvenient...
Posted it here more as a way to document the behavior.

> First thought that came to mind to opt-out is allowing collaborators to add a special kind of label to the PR, e.g. `bot-no-ci-comment` or similar 🤷‍♂️

Sound great.

---

> what you mean by "resume"

For CI jobs with flakes we use:
image
which runs only the failed parts.

PR #226, opened 2019-03-26
Edit CI comment when subsequent PR jobs are started

phillipj:
The motivation behind these changes are to prevent flooding PRs with CI comments from the bot whenever new node-test-pull-request jobs are started. Instead if the bot has already posted a CI comment earlier, it will update the last CI comment and change the URL in that comment to the new job that got started.

Snook in a bugfix for the already existing helper function that fetches the bot's username. There has obviously been a regression in that function, so replaced the existing zero-value-unit-test with an integration test that should increase its reliability going forward.

/cc @nodejs/github-bot @cjihrig


lib/push-jenkins-update.js
```diff
@@ -118,6 +132,25 @@ function createGhStatus (options, logger, cb) {
   })
 }
 
+function findExistingCiComment (options, logger, cb) {
+  async.parallel([
+    botUsername.resolve,
```
refack: This can be a constant, it's not going to change, and it's costing us and API call.

cjihrig:
I haven't reviewed the code, but I'm very 👍 on this idea, and thank you for the quick turnaround time on it.

> Instead if the bot has already posted a CI comment earlier, it will update the last CI comment and change the URL in that comment to the new job that got started.

Does that include the "automatic CI" comment? I think that would be ideal so that there is one comment at the top of the issue that can be treated as an audit log of sorts. This would also mean appending the comment and not just overwriting it.

lib/push-jenkins-update.js
```diff
@@ -118,6 +132,25 @@ function createGhStatus (options, logger, cb) {
...
+    githubClient.issues.getComments.bind(githubClient.issues, {
```
refack: This will only return one page (i.e. 30), we have issues (https://github.com/nodejs/node/pull/26322) with much more, so this won't find the comment and revert to the current behavior...

I'm playing with the graphQL API, that can return 100. I was thinking of fetching a batch from the top, and from the bottom.

BTW: the graphQL schema has `viewerDidAuthor` property, so no need for the bot's id anyway.

lib/push-jenkins-update.js
```diff
@@ -15,7 +17,19 @@ function pushStarted (options, build, cb) {
...
+      if (err) {
+        logger.error(err, 'Got error while retrieving GitHub comments for PR')
+        return
+      }
+
+      const body = `CI: ${build.url}`
+      if (existingComment === undefined) {
+        createPrComment(Object.assign({ number: pr }, options), body)
+      } else {
+        editPrComment(options, existingComment.id, body)
```
refack: IMO it should be 
```suggestion
        editPrComment(options, existingComment.id, `${existingComment.body}\n${body}`)
```

refack:
> Does that include the "automatic CI" comment?

I think it should. It's a convenient place, and it's easy to find (see paging) for PRs authored by Collaborators.

lib/push-jenkins-update.js
phillipj: Good catch, obviously didn't remember that at first. Fetching all pages of entries is something we'd have to do elsewhere, e.g. in [./lib/node-repo.js](https://github.com/nodejs/github-bot/blob/master/lib/node-repo.js#L138).

> I'm playing with the graphQL API, that can return 100. I was thinking of fetching a batch from the top, and from the bottom.

Cool, the GraphQL API is something I've been wanting to try out for a while too!

lib/push-jenkins-update.js
phillipj: Great suggestion 👍

lib/push-jenkins-update.js
phillipj: True enough.

Finding the bot's username via the GitHub API was made with a good intention a while back, but I won't argue with you it surely smells like overengineering.

lib/push-jenkins-update.js
refack: I'll try to push a suggestion, as soon as I refresh my sinon & supertest skills...

phillipj:
Thanks for the lightning quick reviews!

> @cjihrig said:
> Does that include the "automatic CI" comment?

My first intention was to put out the immediate "fire" of flooding CI comments since yesterday (https://github.com/nodejs/github-bot/pull/224). On second thought it probably won't be many changes needed to what is currently in this PR to make that happen, and thereby fix both while we're at it.

> @refack said:
> Similar to the implementation I made.

Mind opening a PR with what you've got if it's ready-ish? I'm certain we'd end up with a better solution at the end if we share notes so to speak..

lib/push-jenkins-update.js
phillipj: https://github.com/nodejs/github-bot/pull/227

refack:
> Mind opening a PR with what you've got if it's ready-ish? I'm certain we'd end up with a better solution at the end if we share notes so to speak..

https://github.com/nodejs/github-bot/pull/228

sam-github:
Perhaps I'm in the minority, but I love having a new CI commit for every CI run. Currently, I manually paste the CI commit in for *every single run* I do, so the existing behaviour just automates what I do manually.

phillipj:
@sam-github in general I feel the same way as you do, but I'm not an active core contributor these days and have never been involved in many PRs simultaneously so 🤷‍♂️

phillipj:
Rebased and force pushed after the bot username got hardcoded in https://github.com/nodejs/github-bot/commit/4f4ac439de4e094490e2009418685aea42989b83, thereby removing the previous complexity of getting the username async from the API.

phillipj:
Closing this in favour of https://github.com/nodejs/github-bot/pull/228 which is more in line with keeping *one* bot comment that is edited as soon as new builds are kicked off, and gives GitHub's GraphQL API a shot which is valuable getting some experience with for later.

PR #230, opened 2019-03-29
Reduce parameter for ci job

refack:
This should work in any case, since we pass the defaults anyway

Also:
* Trigger bot note for `Lite-CI`


refack:
It's working (and generating a simpler URL that's consistent with the new comment generated for full CI runs)
image

PR #244, opened 2019-09-21
Fix to use the port value of dotenv

vomvoru:
[As-is]
It not use the port value of dotenv.

[To-be]
It use the port value of dotenv.


phillipj:
Thanks a lot for contributing, good catch! 💯

PR #250, opened 2020-01-06
labels: add "wasi" label

ZYSzys:


test/unit/node-labels.test.js
```diff
@@ -203,6 +203,16 @@ tap.test('label: "libuv" when ./deps/uv/ files has been changed', (t) => {
   t.end()
 })
 
+tap.test('label: "wasi" when ./deps/uv/ files has been changed', (t) => {
```
phillipj: Should be `./deps/uvwasi/`?

ZYSzys: Fixed, Thanks !

PR #265, opened 2020-07-01
feat: ping reviewers based on which files were changed

mmarchini:
Using a custom OWNERS file, `github-bot` will ping the appropriate teams
based on which files were changed in a Pull Request. This feature is
inteded to work around GitHub's limitation which prevents teams without
explicit write access from being added as reviewers (thus preventing the
vast majority of teams in the org from being used on GitHub's CODEOWNERS
feature).

The OWNERS file is a yaml file (to simplify parsing) composed of
key-value paris. The key is always a string and represents a glob of the
changed files. The value is always an array of strings, and each string
is a team to be pinged.

Ref: https://github.com/nodejs/node/issues/33984
Ref: https://github.com/nodejs/node/pull/34150


mmarchini:
Still need to add tests, but I'm looking for feedback on the implementation (using a YAML file, fetching from core, etc.). This has been discussed already in https://github.com/nodejs/node/pull/33895 and https://github.com/nodejs/node/issues/33984 (and a couple old pull requests where CODEOWNERS were first introduced but then removed because of the GitHub limitations mentioned above).

phillipj:
Conceptually looks good to me 👍

The fact that the definition lives inside the node repo is surely a step in a more sustainable direction going forward.

I'm not opinionated about the format and naming of `OWNERS | CODEOWNERS`, whatever makes most sense for the core maintainers.

lib/node-repo.js
```diff
@@ -1,42 +1,88 @@
...
+      return options.logger.error(err, 'Couldn\' determine default branch')
+    }
+
+    const { default_branch: defaultBranch } = data
+
+    const url = `https://raw.githubusercontent.com/${owner}/${repo}/${defaultBranch}/OWNERS.yml`
+    debug(`Fetching OWNERS on ${url}`)
+    request(url, (err, res, body) => {
+      if (err || !res || res.statusCode >= 400) {
+        return options.logger.error(err, 'Error retrieving OWNERS')
```
lundibundi, line 62: Perhaps, also log `res` (or at least `res.statusCode`) in this case?

mmarchini:
Previous version was a merge-conflict show, it was easier for me to go revert the changes and copy them manually. I added all new functions after current function to avoid any future conflicts if this doesn't land soon.

Main changes:

  - Rewrote everything in `async/await` fashion since we started to use it
  - Broke down all steps into separate function to make testing easier
  - The format is not YAML anymore, we're using the same format GitHub does
    - No need to create a new file on nodejs/node amymore, as soon as we land this the bot will start ping folks :D
  - Added tests
    - I used nock for all tests because using sinon to stub github API calls will supress deprecation warnings that are helpful when bumping major versions of the github rest API package

lib/node-repo.js
mmarchini: Agreed, but can be done in a follow up PR (so we don't block the PR)

lib/node-repo.js
```diff
@@ -185,10 +189,115 @@ function stringsInCommon (arr1, arr2) {
...
+    throw err
+  }
+}
+
+async function getDefaultBranch ({ owner, repo, logger }) {
+  try {
+    const data = (await githubClient.repos.get({ owner, repo })).data || { }
+
+    if (!data['default_branch']) {
+      logger.error(null, 'Couldn\' determine default branch')
```
targos, line 223:
```suggestion
      logger.error(null, 'Could not determine default branch')
```

lib/node-repo.js
```diff
@@ -185,10 +189,115 @@ function stringsInCommon (arr1, arr2) {
...
+  })
+}
+
+async function resolveOwnersThenPingPr (options) {
+  const { owner, repo } = options
+  const times = options.retries || 5
+  const interval = options.retryInterval || fiveSeconds
+  const retry = fn => Aigle.retry({ times, interval }, fn)
+
+  options.logger.debug('getting file paths')
```
targos, line 252:
```suggestion
  options.logger.debug('Getting file paths')
```

lib/node-repo.js
```diff
@@ -185,10 +189,115 @@ function stringsInCommon (arr1, arr2) {
...
+  options.number = options.prId
+  const filepathsChanged = await retry(() => getFiles(options))
+
+  options.logger.debug('getting default branch')
```
targos, line 256:
```suggestion
  options.logger.debug('Getting default branch')
```

lib/node-repo.js
```diff
@@ -185,10 +189,115 @@ function stringsInCommon (arr1, arr2) {
...
+  const defaultBranch = await retry(() => getDefaultBranch(options))
+
+  const url = getCodeOwnersUrl(owner, repo, defaultBranch)
+  options.logger.debug(`Fetching OWNERS on ${url}`)
+
+  const file = await retry(() => getCodeOwnersFile(url, options))
+
+  options.logger.debug('parsing codeowners file')
```
targos, line 264:
```suggestion
  options.logger.debug('Parsing codeowners file')
```

lib/node-repo.js
```diff
@@ -185,10 +189,115 @@ function stringsInCommon (arr1, arr2) {
...
+  const owners = Owners.fromFile(file)
+  const selectedOwners = owners.getOwnersForPaths(filepathsChanged)
+
+  options.logger.debug('pinging codeowners file')
```
targos, line 268:
```suggestion
  options.logger.debug('Pinging codeowners file')
```

mmarchini:
@phillipj do you want to take another look before we land this? Was hoping to land it this week.

mmarchini:
@phillipj feel free to ping me if you have any comments, suggestions or concerns on this PR. I can send a patch addressing those or we can revert if necessary.

phillipj:
Gotcha! Sorry for the late reply, your ping a few days ago drowned in my
mailbox somehow.

PR #267, opened 2020-07-22
label: mark all quic change as `dont-land-v14.x`

MylesBorins:
As per discussion on GitHub it does not appear that QUIC is intended
to be backported to v14.x, add appropriate label.

/cc @jasnell to confirm


richardlau:
@MylesBorins That also implies we should add the other dont-land-* labels (e.g. `dont-land-v12.x`)?

MylesBorins:
I've added the `dont-land-v12.x` label as well. Did not include 10.x as it is in maintenance

MylesBorins:
@targos PTAL

MylesBorins:
is there anything else we need to do to deploy this? Do we have CD in place?

phillipj:
CD is setup, yes. The bot automatically gets deployed when changes lands in master 🚀

PR #268, opened 2020-07-31
chore(lite-ci): disable lite-ci

mmarchini:
Removes minimum amount of code necessary to disable Lite-CI from
triggering (while also passing tests and linters). Lite-CI doesn't work
anymore and now we have Actions testing on multiple platforms and
running linters, therefore we also don't need Lite-CI
anymore. There might be some leftover code related to Lite-CI, but this
fixes the immediate issue.


phillipj:
@nodejs/citgm are you folks using @nodejs-github-bot to kick off CI with a comment anymore?

Asking cause deleting `scripts/trigger-jenkins-build.js` will also remove that initial CITGM effort (https://github.com/nodejs/github-bot/pull/128), not just Lite-CI in nodejs/node upon PRs opened.

mmarchini:
I never started CITGM via comment, had no idea it was an option, and never saw anyone doing it :)

richardlau:
FWIW I made a small fix to the lite-ci pipeline: https://github.com/nodejs/node/pull/34531#issuecomment-667929763

re. citgm, I can't remember the last time I used that feature, or if it still works. (Or, for that matter, whether it triggered the job to test CITGM itself or the job to test lookup.json changes.) If we do want that feature I've a soft preference to adopting the add a label to start CI mechanism that was recently added for core.

mmarchini:
> FWIW I made a small fix to the lite-ci pipeline: nodejs/node#34531 (comment)

I think we should just remove lite-ci from Jenkins. Since we started adding more Actions it doesn't bring any value to us and it ends up being one more Jenkins job to maintain (I'm not even sure if it is working, the last run took only five minutes).

richardlau:
> 
> 
> > FWIW I made a small fix to the lite-ci pipeline: [nodejs/node#34531](https://github.com/nodejs/node/pull/34531) (comment)
> 
> I think we should just remove lite-ci from Jenkins. Since we started adding more Actions it doesn't bring any value to us and it ends up being one more Jenkins job to maintain (I'm not even sure if it is working, the last run took only five minutes).

I have no objections. I think the only reasons people run lite-ci is they're either not up to date with policy (just Actions are adequate for doc-only PRs) or are trying to silence `git node land ...` (which itself hasn't caught up with policy for doc-only PRs when it checks for CI runs).

(five minutes is plausible, the lite job only runs the linters and build/test on LinuxONE which can complete that quickly if there are no code changes and a populated ccache.)

mmarchini:
I'll try to get ncu to recognize Actions as well (we should be requiring it to succeed anyway since we have tests that only run there)

PR #270, opened 2020-08-02
chore(deps): replace `github` with `@octokit/rest`

mmarchini:
`github` has been renamed to `@octokit/rest`. Also bumped to ^v15 as there were no significant breaking changes before ^v16. Upgrading to v16 will require more effort on our side, including:

  - Change all GitHub API calls from callback-style to Promise-style
  - Rework tests so we don't use sinon to replace GitHub methods as sinon will prevents GitHub deprecation warnings. Use nock instead
  - Increase test coverage to include all GitHub API calls
  - Fix deprecation warnings (preferably one by one so it's easier to revert if the bot breaks)

Ref: https://github.com/octokit/rest.js/releases/tag/v14.0.0
Ref: https://github.com/octokit/rest.js/releases/tag/v15.0.1

### DIFF
--- a/lib/github-client.js
+++ b/lib/github-client.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const GitHub = require('github')
+const GitHub = require('@octokit/rest')
 
 const githubClient = new GitHub({
   version: '3.0.0',
--- a/package-lock.json
+++ b/package-lock.json
@@ -151,15 +151,18 @@
       }
     },
     "@octokit/rest": {
-      "version": "14.0.9",
-      "resolved": "https://registry.npmjs.org/@octokit/rest/-/rest-14.0.9.tgz",
-      "integrity": "sha512-irP9phKfTXEZIcW2R+VNCtGHZJrXMWmSYp6RRfFn4BtAqtDRXF5z9JxCEQlAhNBf6X1koNi5k49tIAAAEJNlVQ==",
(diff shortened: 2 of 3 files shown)


phillipj:
FYI had my hopes this update would get rid of the deprecation email we've been getting to the @nodejs-github-bot' email, unfortunately not;

```
Hi @nodejs-github-bot,

On August 5th, 2020 at 04:53 (UTC) your personal access token (General github-bot token) using Node.js GitHub Bot v1.0-beta octokit.js/15.18.3 was used as part of a query parameter to access an endpoint through the GitHub API:

https://api.github.com/repositories/27193779/pulls/34629/files

Please use the Authorization HTTP header instead, as using the `access_token` query parameter is deprecated. If this token is being used by an app you don't have control over, be aware that it may stop working as a result of this deprecation.

Depending on your API usage, we'll be sending you this email reminder on a monthly basis for each token and User-Agent used in API calls made on your behalf.
Just one URL that was accessed with a token and User-Agent combination will be listed in the email reminder, not all.

Visit https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param for more information about suggested workarounds and removal dates.

Thanks,
The GitHub Team
```

mmarchini:
That'll take some more upgrades (which is on my list :) )

phillipj:
Goodie, no worries 👍

PR #271, opened 2020-08-02
chore(jenkins): refactor jenkins-status using events

mmarchini:


test/integration/push-jenkins-update.test.js
```diff
@@ -14,74 +14,93 @@ require('../../scripts/jenkins-status')(app, events)
...
+    .on('replied', (req, interceptor) => {
+      t.doesNotThrow(() => interceptor.scope.done())
+    })
+  nock('https://api.github.com')
     .filteringPath(ignoreQueryParams)
     .post('/repos/nodejs/citgm/statuses/8a5fec2a6bade91e544a30314d7cf21f8a200de1')
     .reply(201)
+    .on('replied', (req, interceptor) => {
+      t.doesNotThrow(() => interceptor.scope.done())
+    })
```
phillipj, line 53: Interesting alternative approach to verifying scopes that I haven't used before. Any obvious upsides compared to the previous approach?

test/integration/push-jenkins-update.test.js
```diff
@@ -14,74 +14,93 @@ require('../../scripts/jenkins-status')(app, events)
...
   supertest(app)
     .post('/citgm/jenkins/start')
     .send(jenkinsPayload)
-    .expect(201)
+    .expect(200)
     .end((err, res) => {
-      prCommitsScope.done()
-      scope.done()
+      // prCommitsScope.done()
+      // scope.done()
```
phillipj, line 63: Assuming we don't need this commented lines anymore.

phillipj:
Nice refactor 💯

test/integration/push-jenkins-update.test.js
mmarchini: Now that we're using EventEmitter and Promises, the timing is different and as a result the callback below to run before we made a call to the `scope` URL, so we get an error when we call `scope.done()`

phillipj: Ahh okey, so that's another approach to fixing what was my biggest headache in https://github.com/nodejs/github-bot/pull/258 basically.

What I ended up doing there, was to make sure all event handlers returned a `Promise`. That way the resulting `Promise` coming out of the `app.emitJenkinsEvent()`-equivalent of https://github.com/nodejs/github-bot/pull/258 would actually resolve after all handlers had completed. In the end we can be sure that all handlers (and thereby `scope`) has finished when the request gets the 200 response back.

No biggie tho, just wanted to understand the pros & reasons behind this new approach introduced.

mmarchini: That works too, but no reason to delay the response until app.emitJenkinsEvent finishes since Jenkins doesn't do anything with 200/500 status codes. We could for testability if we wanted to, but this approach seems fine and more robust?

phillipj:
I'm not opinionated about how it's done, as long as we can trust the test
results I'm all in.

PR #275, opened 2020-08-08
fix(owners): path shouldn't start with .

mmarchini:


mmarchini:
cc @nodejs/github-bot

lib/node-owners.js
```diff
@@ -1,9 +1,6 @@
 'use static'
 
-const { parse } = require('codeowners-utils')
-// NOTE(mmarchini): `codeowners-utils` doesn't respect ./ prefix,
-// so we use micromatch
-const micromatch = require('micromatch')
```
phillipj, line 6: This can be uninstalled now, right?

mmarchini: Right, forgot to uninstall before committing. Will do it before landing

PR #277, opened 2020-08-10
chore: dont use sinon on push-jenkins-update test

mmarchini:
`@octokit/rest` uses deprecation warning to indicate breaking changes,
and sinon suppresses those warnings when running tests. Replacing sinon
with nock for testing ensures we fix all deprecation warnings before
bumping octokit. This commit replaces sinon with nock in one test.


test/common.js
```diff
@@ -0,0 +1,11 @@
...
+
+const url = require('url')
+
+function ignoreQueryParams (pathAndQuery) {
+  return url.parse(pathAndQuery, true).pathname
+}
+
+module.exports = {
+  ignoreQueryParams
+}
```
phillipj, line 11: 👍

phillipj:
> Replacing sinon with nock for testing ensures we fix all deprecation warnings before bumping octokit.

Nice catch that I hadn't considered.

PR #281, opened 2020-08-12
chore: use listEvents instead of getEvents

mmarchini:
getEvents is deprecated. Also use `number` instead of `issue_number` as
argument (`issue_number` is deprecated).


mmarchini:
Depends on #280 (otherwise I'd need to rewrite all sinon calls, which would generate a lot of merge conflicts)

PR #272, opened 2020-08-02
feat: relay jenkins and gh events to gh

mmarchini:
Initial implementation of the "Actions relay" suggested here: https://github.com/nodejs/github-bot/issues/264. Doesn't work yet because createDispatchEvent is not available on the `github` version we're using.

(current PR rebased on top of #271 and #270, actual implementation [here](https://github.com/nodejs/github-bot/pull/272/commits/4907f69e6311cb75b774f9faba12c2486125dded))

PR #282, opened 2020-08-15
feat: upgrade octokit to 16

mmarchini:
Resolved all remaining deprecations on v15 (convert everything to promises because callbacks were removed), upgraded to v16, fixed tests (the payload format of one request changed and nock was not matching it anymore). I already fixed some deprecations on v16, but didn't commit yet so that the PR is not overwhelming. With this PR we'll have access to [createDispatchEvent](https://github.com/octokit/rest.js/releases/tag/v16.29.0), which means it will unblock #272.

P.S.: Each commit converts one callback-based function into Promises/async, so the best way to read this PR is going on each commit individually.

### DIFF
--- a/lib/node-repo.js
+++ b/lib/node-repo.js
@@ -1,7 +1,6 @@
 'use strict'
 
 const LRU = require('lru-cache')
-const retry = require('async').retry
 const Aigle = require('aigle')
 const request = require('request')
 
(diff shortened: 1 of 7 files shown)


lib/push-jenkins-update.js
```diff
@@ -29,13 +29,10 @@ function pushStarted (options, build, cb) {
...
-    if (err) {
-      logger.error(err, 'Got error when retrieving GitHub commits for PR')
-      cb(err)
-      return
-    }
-
+  findLatestCommitInPr(optsWithPr).catch(err => {
+    logger.error(err, 'Got error when retrieving GitHub commits for PR')
+    return cb(err)
+  }).then(latestCommit => {
```
targos, line 35: I think chaining like this will still call the `then` callback even in case of an error. To avoid that, you can use `findLatestCommitInPr(optsWithPr).then(onSuccess, onError)`

lib/push-jenkins-update.js
```diff
@@ -57,13 +54,10 @@ function pushEnded (options, build, cb) {
...
```
targos, line 60: Ditto

lib/node-repo.js
```diff
@@ -74,29 +74,32 @@ async function updatePrWithLabels (options, labels) {
...
+      number: options.prId,
+      name: label
+    })
+  } catch (err) {
+    if (err.code === 404) {
+      options.logger.info('Label to remove did not exist, bailing ' + label)
+      return
     }
+    options.logger.error(err, 'Error while removing a label')
+    return
```
targos, line 98:
```suggestion
    throw err
```

lib/node-repo.js
```diff
@@ -149,11 +149,7 @@ function getBotPrLabels (options, cb) {
     page: 1,
     per_page: 100, // we probably won't hit this
     number: options.prId
-  }, (err, res) => {
-    if (err) {
-      return cb(err)
-    }
-
+  }).catch(cb).then(res => {
```
targos, line 152: Ditto: `.then(res => { ... }, cb)` or `.then(res => { ... }).catch(cb)` in case an error happens in the success callback

lib/node-repo.js
mmarchini: Would probably break the backport stuff because this code path is not well tested

lib/push-jenkins-update.js
mmarchini: Humm, interesting, I guess that makes sense. Will test tomorrow and update the code if that's the case.

lib/node-repo.js
targos: It was calling `cb(err, null)` in that case. Isn't throwing the equivalent in an async function?

mmarchini: It wasn't, the function didn't take a callback, I added the callback on the first commit to test it a little bit better.

This is how it is called today on master: https://github.com/nodejs/github-bot/blob/c90f7c18742be6566cba666048b5c16c4c1842d0/scripts/attempt-backport.js#L143

On a second thought, I guess throwing here wouldn't break anything, would just leave the rejection unhandled (which we can handle in the script). It still worries me a bit because this function isn't called as part of the backport tests, but I guess it should be fine. Will change it tomorrow to throw

mmarchini:
@targos I think I addressed everything :)

targos:
LGTM

PR #287, opened 2020-11-13
Replace Travis CI with GitHub Actions + more Node.js versions

phillipj:
Primarily as a precursor to upgrading the Node.js version on the server where @nodejs-github-bot runs from v10 -> v14, running tests on v14 in CI is valuable.

Also added Node.js 15 to the mix as it's worth getting early heads up if there are any upcoming issues before considering an upgrade.

The transition to GitHub Actions is done because it's convenient and there's been significant use of Actions in the Node.js org lately.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,20 @@
+name: Test
+
+on: [push, pull_request]
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [10.x, 12.x, 14.x, 15.x]
+
+    steps:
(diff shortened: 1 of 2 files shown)


richardlau (approved):

mmarchini (approved):

richardlau:
> The transition to GitHub Actions is done because it's convenient and there's been significant use of Actions in the Node.js org lately.

Also Travis have recently reduced allowances: https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

phillipj:
Appreciate the lightning quick reviews!

> Also Travis have recently reduced allowances

Huh, thanks for sharing, I wasn't aware of that.

PR #118, opened 2017-01-27
attempt-backport: temporary disable dont-land-on-* labels

phillipj:
There's been some concerns about auto labelling `dont-land-on-*` labels lately, especially since the bot is often wrongly adding these labels ATM. Therefore temporarily disabling that functionality to avoid further damage.

Refs https://github.com/nodejs/github-bot/issues/116

PR #288, opened 2020-11-15
Remove unused attempt backport script

phillipj:
This was a very neat part of the bot at one point (https://github.com/nodejs/github-bot/issues/77), but ended up creating some noise and parts got disabled (https://github.com/nodejs/github-bot/pull/118).

It used to try backport of patches from created & updated PRs against the `${version}.x-staging` branches and as a result, added `lts-watch-v${version}.x` or `dont-land-on-v${version}.x` when applicable.

When it was operational, it tried to apply PR patches against Node.js versions: 4, 6 and 7. Those version are all end-of-life now.

For the bot to try backport of patches for newer versions of Node.js, those versions would have to be added explicitly in the bot' source code. That has obviously not happened, assuming the reason being it wasn't as helpful afterall.


richardlau:
I'd completely forgotten we'd tried this before. Not sure if the other current members of @nodejs/releasers are aware of this either/think this would be useful to resurrect in an updated form (maybe instead of `lts-watch-*` labels it could focus on `backport-requested-*`?).

I have a vague recollection that @codebytere may have mentioned that Electron has a bot that does something similar?

phillipj:
@nodejs/releasers you think it would be useful this resurrect this, ref @richardlau's comment above?

If not, I'd really appreciate if we could sunset this code for now. We can always dig it back up again later if needed..

phillipj:
Since no objections were made, I'm merging this to remove this unused part of the source code for now.

PR #309, opened 2022-01-12
Old function convert in arrow function

sharozraees802:


phillipj:
Hi @sharozraees802,

thanks for your contribution!

Mind elaborating **why** you would like to see this change land? What benefits does it bring to the project?

PR #312, opened 2022-02-15
Switch CI to pkgjs/action

dominykas:
Closes https://github.com/nodejs/github-bot/issues/310

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "test:watch": "nodemon -q -x 'npm test'"
   },
   "engines": {
-    "node": ">= 6.0.0"
+    "node": ">= 8.0.0"
   },
   "private": true,
   "license": "MIT",
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,17 +4,4 @@ on: [push, pull_request]
 
 jobs:
   tests:
-    runs-on: ubuntu-latest
-
-    strategy:
-      matrix:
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
     "test:watch": "nodemon -q -x 'npm test'"
   },
   "engines": {
-    "node": ">= 6.0.0"
+    "node": ">= 8.0.0"
```
dominykas, line 11: Technically, this is a breaking change, however the previous CI had already bumped to test in node@10+. 8 is still passing, but maybe I should bump it to something even more modern?

I figure node@6 might have been only failing because it is using npm@3 by default, which does not yet have `package-lock.json`, so I could try to address that in `pkgjs/action` (by installing newer npm), and then we could try reverting this change for supported node?

LMK what's the preferred course.

phillipj: No worries. Guess we've just forgotten to bump this when we've upgraded on the github-bot server.

ATM we're running Node.js 14, so although this version range isn't necessarily wrong, it's surely misleading.

> .. but maybe I should bump it to something even more modern?

Yeah, sounds like a good idea to me. Having the above in mind, having `>= 14.0.0` would make sense, right?

phillipj:
Very cool indeed, great initiative 👏🏻

phillipj (approved):
LGTM.

Feel free to do what's most sensible to you with regards to `engines.node`.

Issue #376, opened 2023-03-07
Bot failing to start: `TypeError: glob.sync is not a function`


richardlau:
@tniessen [reports](https://openjs-foundation.slack.com/archives/C03BJP63CH0/p1678138229956679?thread_ts=1678104091.867669&cid=C03BJP63CH0) that Jenkins status is still not being posted to pull requests. 

Looking at https://ci.nodejs.org/job/post-build-status-update runs, e.g. https://ci.nodejs.org/job/post-build-status-update/1837783/flowGraphTable/, I can see that although the jobs are green, the HTTP request to the bot has failed:
![image](https://user-images.githubusercontent.com/5445507/223419928-cb5b0042-2c89-4ff5-9c59-cad0e69eb2dc.png)

https://ci.nodejs.org/job/post-build-status-update/1837783/execution/node/17/
```console
hudson.AbortException: Fail: Status code 408 is not in the accepted range: 100:399
```

If I log into the server for the github-bot, I can see that the service isn't running:
```console
root@infra-rackspace-debian8-x64-1:~# systemctl status github-bot
● github-bot.service - github-bot
   Loaded: loaded (/lib/systemd/system/github-bot.service; enabled)
   Active: activating (auto-restart) (Result: exit-code) since Tue 2023-03-07 12:19:55 UTC; 41ms ago
  Process: 25524 ExecStart=/usr/bin/node server.js (code=exited, status=1/FAILURE)
 Main PID: 25524 (code=exited, status=1/FAILURE)

Mar 07 12:19:55 infra-rackspace-debian8-x64-1 systemd[1]: github-bot.service: main process exited, code=exited, status=1/FAILURE
Mar 07 12:19:55 infra-rackspace-debian8-x64-1 systemd[1]: Unit github-bot.service entered failed state.
root@infra-rackspace-debian8-x64-1:~#
```

and from the syslog:

```console
Mar  7 12:13:26 infra-rackspace-debian8-x64-1 systemd[1]: Started github-bot.
Mar  7 12:13:26 infra-rackspace-debian8-x64-1 node[24754]: /home/iojs/github-bot/server.js:13
Mar  7 12:13:26 infra-rackspace-debian8-x64-1 node[24754]: glob.sync(scriptsToLoad).forEach((file) => {
Mar  7 12:13:26 infra-rackspace-debian8-x64-1 node[24754]: ^
[...]

PR #377, opened 2023-03-07
fix: replace `glob.sync` for glob@9

richardlau:
glob@9 no longer exports `sync()`.

Fixes: https://github.com/nodejs/github-bot/issues/376
Refs: https://github.com/isaacs/node-glob/issues/493
Refs: https://github.com/nodejs/github-bot/pull/372

### DIFF
--- a/server.js
+++ b/server.js
@@ -2,7 +2,7 @@
 
 require('dotenv').config({ silent: true })
 
-const glob = require('glob')
+const { globSync } = require('glob')
 const logger = require('./lib/logger')
 
 const port = process.env.PORT || 3000
@@ -10,7 +10,7 @@ const scriptsToLoad = process.env.SCRIPTS || './scripts/**/*.js'
 const { app, events } = require('./app')
 
 // load all the files in the scripts folder
-glob.sync(scriptsToLoad).forEach((file) => {
+globSync(scriptsToLoad).forEach((file) => {
   logger.info('Loading:', file)
   require(file)(app, events)
 })


richardlau:
Alternatively we could revert https://github.com/nodejs/github-bot/pull/372.
According to https://github.com/isaacs/node-glob/blob/main/changelog.md#90, glob@9 drops support for Node.js versions earlier than 16 and we're currently running the bot on Node.js 14:
```console
# node -v
v14.16.1
#
```

targos:
I think we should bump the server to at least v16 (ideally v18 to be future-proof).
v14.x is very close to its EoL date.

Trott (approved):

richardlau:
> I think we should bump the server to at least v16 (ideally v18 to be future-proof). v14.x is very close to its EoL date.

Tried. Blocked on the server still running Debian 8 -- we're using Nodesource distributions of Node.js on the server and they don't have newer Node.js packages for Debian 8 (Jessie). https://github.com/nodejs/build/issues/3214

Trott:
Should we go ahead and land this and hope it fixes the current issue?

Trott:
Hmmm...it looks like the actions are green when they are failing... https://github.com/nodejs/github-bot/actions/runs/4354160252/jobs/7609125321#step:12:1467 fails but is green....

richardlau:
> Should we go ahead and land this and hope it fixes the current issue?

I'll go ahead and merge and see if the bot at least starts. It can't be worse that the current state, and if it doesn't we can revert this and the glob update temporarily while we get the server updated.

PR #378, opened 2023-03-07
fix: add leading . for paths loaded via require

Trott:

PR #379, opened 2023-03-07
fix: use `dotRelative` glob option

richardlau:
Refs: https://github.com/nodejs/github-bot/issues/376#issuecomment-1458553582
Refs: https://github.com/isaacs/node-glob/issues/495

### DIFF
--- a/server.js
+++ b/server.js
@@ -10,9 +10,9 @@ const scriptsToLoad = process.env.SCRIPTS || './scripts/**/*.js'
 const { app, events } = require('./app')
 
 // load all the files in the scripts folder
-globSync(scriptsToLoad).forEach((file) => {
+globSync(scriptsToLoad, { dotRelative: true }).forEach((file) => {
   logger.info('Loading:', file)
-  require(`./${file}`)(app, events)
+  require(file)(app, events)
 })
 
 app.listen(port, () => {


richardlau:
Difference with https://github.com/nodejs/github-bot/pull/378 is if absolute paths are used in `process.env.SCRIPTS` (which AFAICT we are not setting on the server) -- https://github.com/nodejs/github-bot/pull/378 will always append the leading `./`.
https://github.com/nodejs/github-bot/blob/d28e26066524e70c050469969b03a911e387dd6d/server.js#L9

server.js
```diff
@@ -10,7 +10,7 @@ const scriptsToLoad = process.env.SCRIPTS || './scripts/**/*.js'
 const { app, events } = require('./app')
 
 // load all the files in the scripts folder
-globSync(scriptsToLoad).forEach((file) => {
+globSync(scriptsToLoad, { dotRelative: true }).forEach((file) => {
   logger.info('Loading:', file)
   require(file)(app, events)
```
Trott, line 15: I don't know why it's not showing up in this diff, but this line is currently:

```
require(`./${file}`)(app, events)
```

and will need to be adjusted.

Trott: Maybe rebase the pr branch against main?

richardlau: Rebased and updated.

Trott (approved):

Trott:
Tested locally (since server.js isn't covered by the tests) and it works.

PR #383, opened 2023-03-17
 chore: update @octokit/rest to 19.0.7

Trott:
The underlying implementation changed and it now uses fetch, which nock  cannot mock. So introduce fetch-mock and change tests where relevant.


phillipj:
Impressive @Trott! 👏🏻

PR #511, opened 2024-05-14
fix: pin glob to 10.3.4

Trott:
We can unpin it once the bot is running on Node.js 14.18 or newer.

Fixes: https://github.com/nodejs/build/issues/3720

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "dotenv": "^16.4.5",
     "events-async": "^1.2.1",
     "express": "^4.19.2",
-    "glob": "^10.3.15"
+    "glob": "10.3.14"
   },
   "devDependencies": {
     "eventsource": "^2.0.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
         "dotenv": "^16.4.5",
         "events-async": "^1.2.1",
         "express": "^4.19.2",
-        "glob": "^10.3.15"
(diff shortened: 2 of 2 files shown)


richardlau (approved):

package.json
```diff
@@ -24,7 +24,7 @@
     "dotenv": "^16.4.5",
     "events-async": "^1.2.1",
     "express": "^4.19.2",
-    "glob": "^10.3.15"
+    "glob": "10.3.14"
```
richardlau, line 27: Apparently I fail at reading... was this supposed to be 10.3.4 like the PR/commit title?

Trott: No, but we could try that. I chose 10.3.14 because 10.3.15 is the one where the ["requires 14.18 or newer" commit](https://github.com/isaacs/node-glob/commit/57f5551d92d8bc05a903bcc2601045e4f5b7fa5a) first appears.

Trott: I typo'ed the commit message (and the PR title was taken from the commit message). The previous version we shipped with was 10.3.12, so maybe we can try that.

Trott: The require-a-newer-version-of-14.x commit was the first thing landed after the 10.3.13 release, further suggesting that 10.3.12 is the way to go.

Trott: https://github.com/nodejs/github-bot/pull/512

PR #521, opened 2024-06-05
migrate request reviewers to github actions

RedYetiDev:
This PR is the second half of <https://github.com/nodejs/node/pull/53149> (and it can't be merged until after that). This removes the logic for requesting reviewers from the Github bot, as the functionality (if <https://github.com/nodejs/node/pull/53149> is merged) will be moved to github actions.


Trott:
Since this is dependent on another PR, I'm going to put it into draft mode so no one merges it prematurely.

PR #541, opened 2024-07-22
fixup! build(deps): bump @octokit/rest from 20.1.1 to 21.0.1

Ceres6:
Require of octokit migrated to `await import` to accommodate chango to ESM-only

### DIFF
--- a/lib/github-client.js
+++ b/lib/github-client.js
@@ -1,6 +1,6 @@
 'use strict'
 
-async function getGithubClient() {
+async function getGithubClient () {
   const { Octokit } = await import('@octokit/rest')
 
   const githubClient = new Octokit({
--- a/lib/github-client.js
+++ b/lib/github-client.js
@@ -1,13 +1,17 @@
 'use strict'
 
-const { Octokit } = require('@octokit/rest')
+async function getGithubClient() {
+  const { Octokit } = await import('@octokit/rest')
 
-const githubClient = new Octokit({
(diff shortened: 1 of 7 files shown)


Trott:
Do tests pass for you locally? They're failing in CI....

lib/github-client.js
```diff
@@ -1,13 +1,17 @@
 'use strict'
 
-const { Octokit } = require('@octokit/rest')
+async function getGithubClient() {
```
Trott, line 3:
```suggestion
async function getGithubClient () {
```

Trott:
OK, after a small linter fix, the tests pass with Node.js 22 but not Node.js 20.  Node.js 20 is what we run in production, so we'll want to fix that.

Ceres6:
Seems to be a race condition, so I must have forgotten to await sth. Taking a look rn

Ceres6:
it seems we were missing a couple returns in the original code and tap wasn't waiting for promise resolution. It should be fixed now

Trott:
Tests pass now. 👍 

There is now a potentially-competing PR at https://github.com/nodejs/github-bot/pull/542. If there's benefit to landing this one first as an incremental step, I'm good with that. If not, then I'm really sorry that you put in the work and now this might not land. It happens, but it still sucks, and I'm sorry.

Trott:
Closing in favor of https://github.com/nodejs/github-bot/pull/542. Thank you for the effort, though!

